PR #19, opened 2025-08-21
[MNJPeng] iP

MNJPeng:
# ☠️ **Mael** 

> Ready to ~~take over~~ receive missions

## Purpose
- Load and save missions
- Track completion
- Easy Lookup

## *Mission Variants*
- [x] todo
- [x] deadline
- [x] event
- [ ] recurring event (coming soon)

## Installation Process
1. Download [here](https://github.com/MNJPeng/ip/releases/tag/v0.1)
2. Run java -jar Mael.jar
3. Complete

## Extra modifications
To speed up or disable startup and shutdown sequences, `main` can be modified.

```
public static void main(String[] args) throws InterruptedException {
  new Mael("data/Mael.txt",true, true, null).run();
}
```
- To speed up:
  - Set the first true to false
- To disable:
  - Set the second true to false


meltedham:
💀

src/main/java/mael/taskList/TaskList.java
```diff
@@ -0,0 +1,356 @@
...
+        tasks.add(task);
+        return task.toString();
+    }
+
+    /**
+     * Returns a list of Strings which represent {@code Task} to save
+     * 
+     * @return List of Strings to save in storage
+     */
+    private List<String> getTasks(List<? extends Task> tasks, Function<? super Task, ? extends String> function) {
```
Vorld: Perhaps the name `getTasks` is a little misleading here. It doesn't simply return a list of task objects but in fact also applies a function to each task. Maybe a more descriptive name like `transformTasksToString` would be better here.

src/main/java/mael/Mael.java
```diff
@@ -0,0 +1,65 @@
...
+        this.DATA_PATH_NAME = dataPathName;
+    }
+
+    /**
+     * Runs instance of Mael
+     * 
+     * @param has_delay Enables delay during display 
+     * @param has_sequences Enables launching and closing animations
+     */
+    public void run(boolean has_delay, boolean has_sequences) {
```
Vorld: Perhaps it would be better to follow CamelCase coding style here with the parameters to maintain a consistent coding standard

src/main/java/mael/ui/UI.java
```diff
@@ -0,0 +1,221 @@
...
+     * Displays text when default {@code close} is interrupted
+     */
+    public void safeClose() {
+        System.out.println("\tClose Sequence Interrupted..\n\n\tClosing immediately");
+    }
+
+    /** 
+     * Displays line divider text between inputs and outputs of Mael
+     */ 
+    public void line() {
```
Vorld: Perhaps this can be renamed as verb, `printDividerLine()` so that it is clearer that this actually outputs the divider line in the cli.

It is also a little confusing because line in the next method is referring to text, but here it's referring to a divider.

src/main/java/mael/ui/UI.java
```diff
@@ -0,0 +1,221 @@
...
+        System.out.println("\n" + line);
+    }
+
+    /** 
+     * Displays text line by line with delays
+     * 
+     * @param text Text to be displayed line by line
+     * @throws InterruptedException If sleep is interrupted
+     */
+    private void lineByLine(String text) throws InterruptedException {
```
Vorld: Similar to above, perhaps this can also be renamed as a verb, `printLineByLine(String text)` so it's clearer that this method actually outputs to the CLI

src/main/java/mael/taskList/TaskList.java
```diff
@@ -0,0 +1,356 @@
...
+        return tasks.stream().map(function).collect(Collectors.toList());
+    }
+
+    /**
+     * Returns a list of Strings which represent {@code Task} to print
+     * 
+     * @return List of Strings to print
+     */
+    public List<String> getTasksPrint() {
```
Vorld: For this method and the subsequent `getTasksSave()` method, it can be a little confusing that the return type is a `List<String>`, although admittedly the `Task` class is private. Perhaps it would be clearer to rename as `getTasksAsSaveString()` and `getTasksAsPrintString()`.

Vorld (commented):
Overall, looks mostly great to me!

Just mainly some nitpicks about the naming of some methods, to make their behaviour clearer.

src/main/java/mael/taskList/TaskList.java
MNJPeng: Nice suggestion! Will do!

src/main/java/mael/Mael.java
MNJPeng: Good catch! Thanks!

src/main/java/mael/Mael.java
```diff
@@ -0,0 +1,65 @@
+package mael;
+
+import mael.commands.Command;
+import mael.parser.Parser;
+import mael.storage.Storage;
+import mael.taskList.TaskList;
+import mael.ui.UI;
+
+public class Mael {
```
Chrogain, line 11: Good job! Coding standards looks good to me!

src/main/java/mael/commands/AddCommand.java
```diff
@@ -0,0 +1,76 @@
+package mael.commands;
+
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import mael.MaelException;
```
Chrogain, line 6: There could be a spacing between the import java and import mael.

src/main/java/mael/commands/AddCommand.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Constructor for AddCommand for Event Tasks
+     *
+     * @param title Task title
+     * @param from Start date of Event
+     * @param by End date of Event
+     * @param isCompleted Completion state of Task
+     */
+    public AddCommand(String title, String from, String by, boolean isCompleted, boolean isDisplayed) throws MaelException {
```
Chrogain: Method name can use line wrapping

src/main/java/mael/parser/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+     * Parses inputs from the user in the format required
+     *
+     * @param text User input
+     * @return {@code Command} based on user request
+     * @throws MaelException If user input is not in required format
+     */
+    public static Command parseInput(String text) throws MaelException {
+        String[] sections = text.split(" /");
+        String[] commandSections = sections[0].split(" ", 2);
+        switch (commandSections[0]) {
```
Chrogain, line 36: switch and case have to be on the same indentation

src/main/java/mael/taskList/TaskList.java
```diff
@@ -0,0 +1,368 @@
...
+        tasks.add(task);
+        return task.toString();
+    }
+
+    /**
+     * Returns a list of Strings which represent {@code Task} to save
+     *
+     * @return List of Strings to save in storage
+     */
+    private List<String> getTasksAsStrings(List<? extends Task> tasks, Function<? super Task, ? extends String> function) {
```
Chrogain: Method signature can use line wrapping

src/main/java/mael/taskList/TaskList.java
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Checks date against tasks
+     *
+     * @param dateBy Date to check whether in event durations, or after
+     * deadlines
+     * @return List of Events where {@code dateBy} is in the duration, and
+     * Deadlines where {@code dateBy} is after the deadline
+     */
+    public List<String> checkDate(LocalDateTime dateBy) {
+        return getTasksAsStrings(tasks.stream().filter(task -> task.checkDate(dateBy)).collect(Collectors.toList()), Task::toString);
```
Chrogain: Indentation for the second line should be 8 spaces more than the first line

Chrogain (commented):
Just needs some line wrapping for long names and proper indentation for code longer than 1 line.

PR #30, opened 2025-08-22
[forestmu] iP

forestmu:
# Candy 🍬
> ~Sweet~ task manager Candy

Candy allows you to:
1. **Add** tasks
2. **Delete** tasks
3. **Mark** tasks
4. **Find** tasks

*Examples:*
- [T][ ] Read Book
- [D][ ] Return Book (by: 5 Aug 2025 18:00)
- [E][ ] Reading period (from: 1 Aug 2025 18:00 to: 5 Aug 2025 18:00)

Some commands are:
- `todo`
- `deadline`
- `event`
- `mark`
- `unmark`
- `delete`
- `list`
- `find`

**Include** 
- [ ] task when adding tasks 
- [ ] `/by` for deadline
- [ ] `/from` and `/to` for event
- [ ] number to mark, unmark and delete
- [ ] keyword for the task you are finding

Candy can ***save*** your task! Here's an the `main` method:
```java
        new Candy("./data/candyStorage.txt").run();
    }
```
Find Candy [HERE](https://github.com/forestmu/ip)


src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,162 @@
...
+import storage.Storage;
+import java.util.ArrayList;
+
+/**
+ * Represents all the tasks
+ */
+public class TaskList {
+    private ArrayList<Task> allText;
+    private ArrayList<String> textToSave;
+    private Storage taskStorage;
```
polyrhythmas, line 22: Great job on the naming of variables! They suit the naming conventions.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,162 @@
...
+                + newTask.toString() + "\n    Now you have " + allText.size()
+                + " tasks in your list.";
+    }
+
+
+    /**
+     * Returns string of  all the task found with keyword
+     *
+     * @param keyword string description of what task to find
+     */
```
polyrhythmas: Some minor issues with spacing and capitalising words in the header comment

polyrhythmas (commented):
LGTM! Very well done in terms of OOP encapsulation and following the coding standard.

src/main/java/candy/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+        case EVENT:
+            try {
+                return taskList.addTask(text, "event");
+            } catch (NoEndException | NoStartException
+                     | NoTaskException | InvalidTimeInputException e) {
+                return Ui.printError(e);
+            }
+        }
+        return null;
```
lim-james, line 78: In this method, if a command was introduced in `Command.java` but was not implemented here, the `null` returned will be pass further through the stack into `DialogBox`. Which only has this to handle errors,
```java
} catch (IOException e) {
        e.printStackTrace();
}
```
Though this may seem useful because you're printing the stack trace, but 
1. the methods which could have caused this error might have already been popped off the stack or 
2. would be too verbose to debug a small bug like this

## Consideration
Add a default case to throw a new type of exception

src/main/java/candy/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+    public static String parse(String text, TaskList taskList) {
+        Command commandWord;
+        try {
+            commandWord = Command.fromInput(text);
+        } catch (InvalidInputException e) {
+            return Ui.printError(e);
+        }
+
+        String number;
+        switch (commandWord) {
```
lim-james, line 55: This switch statement works ok now, but if we intend to get more develops on board, would we want all of them to be working on the same switch statement in `Parser::parse`? Consider how git handles changes within the same file, this could result in a bunch of unnecessary merge conflicts down the road.

## Consideration
[Week 3's A-MoreOOP](https://nus-cs2103-ay2526s1.github.io/website/schedule/week3/project.html#a-moreoop)'s stretch goal

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+                String description = parts[2].trim();
+
+                //create the task
+                if (type.equals("T")) {
+                    task = new TodoTask(description, isDone);
+                } else if (type.equals("D")) {
+                    String end = parts[3].trim();
+                    Time endTime = new Time(end);
+                    task = new DeadlineTask(description, isDone, endTime);
+                } else {
```
lim-james: In the earlier two branches we are creating a `TodoTask` for `'T'` and a `DeadlineTask` for `'D'`. However, note that in this branch we are creating an `EventTask` for anything - `'E'`, `'F'`, even `'t'`. Yes, this has no problems when our save files are written by us and no corruption occurs. 

## Consideration
Since you already have an `InvalidInputException` class, you could shift `EventTask` creation into it's proper `'E'` branch and throw that exception under the `else` branch instead.

lim-james (changes requested):
General code quality is good, just a couple of changes that I'd recommend to make the project more airtight and scalable :)

src/main/java/candy/Command.java
```diff
@@ -0,0 +1,43 @@
...
+    Command(String keyword) {
+        this.keyword = keyword;
+    }
+
+    /**
+     * Returns the keyword from text user inputs
+     *
+     * @param input the string of text user input
+     */
+    public static Command fromInput(String input) {
```
Jaey0un: Prefer a verb for method names. Consider parseCommand(String input) instead of fromInput, to clearly communicate the action.

src/main/java/candy/DialogBox.java
```diff
@@ -0,0 +1,72 @@
...
+        dialog.setText(text);
+        displayPicture.setImage(img);
+    }
+
+    /**
+     * Flips the dialog box such that the
+     * ImageView is on the left and text on the right.
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
```
Jaey0un, line 55: Consider a clearer name like childrenReversed or reversedChildren to explain the content.

src/main/java/candy/MainWindow.java
```diff
@@ -0,0 +1,74 @@
...
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+        String candyText = Ui.printWelcome();
+        dialogContainer.getChildren().addAll(
+                DialogBox.getCandyDialog(candyText, candyImage));
+    }
+
+    /**
+     * Injects the Candy instance
+     */
+    public void setCandy(Candy c) {
```
Jaey0un: Use a descriptive parameter name instead of a single-letter. e.g., setCandy(Candy candy).

src/main/java/candy/MainWindow.java
```diff
@@ -0,0 +1,74 @@
...
+    private void handleUserInput() {
+        String userText = userInput.getText();
+        String candyText = candy.getResponse(userInput.getText());
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(userText, userImage),
+                DialogBox.getCandyDialog(candyText, candyImage)
+        );
+        userInput.clear();
+
+        if (Parser.getIsConversationOver()) {
```
Jaey0un, line 72: Consider renaming the boolean accessor to isConversationOver() to follow the boolean naming guideline.

Jaey0un (commented):
Overall, the code is clear and well-structured! It would be better to just avoid single-letter params and use more descriptive names.

PR #33, opened 2025-08-22
[mattcce] iP

mattcce:
# Clanker

Have a clanker manage your tasks for you! It can:
- track your tasks and **completion**
- track **different kinds** of tasks
- understand **dates and times**
- ~fight back~
- _... and more!_

---

> The rest of this is just fluff to demonstrate GFMD usage. I've written more legitimate PRs like [here](https://github.com/source-academy/java-slang/pull/78) and [here](https://github.com/source-academy/java-slang/pull/79). - PR Author 

### More nonsense!

> 

1. Here's a numbered list
2. With a couple items
3. Here's the _third_ item

Leftover tasks:
- [X] Do some reflection 🙏 
- [ ] Do even more reflection 📿 

Proof of reflection:

```java
import java.lang.reflect.Method;

public class Main {
    static class Box<T> {
        T get() { return null; }
        void set(T item) { }
        void doNothing(T item) { }
    }

    static class BoxOfComparable<T extends Comparable<T>> extends Box<T> {
        void set(T item) { }

            BoxOfComparable<?> ok = new BoxOfComparable<>();
            ok.doNothing(null);

            try {
                Class<BoxOfComparable> thisClass = BoxOfComparable.class;
                Method[] methods = thisClass.getDeclaredMethods();

                for (Method method : methods) {
                    System.out.println(method.toString());
                }
            } catch (Throwable e) {
                System.err.println(e);
            }
        }
    }
}
```


src/main/java/fmt/CommandParser.java
```diff
@@ -0,0 +1,78 @@
+package fmt;
```
Javier-Lua: Maybe can use formatter instead of fmt for readability?

src/main/java/clanker/Clanker.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        String deadline = cmd.getOptionValue("by");
+
+        if (deadline == null || deadline.isEmpty()) {
+            writePrompt("Oops! You must specify a deadline!");
+            return;
+        }
+
+        DeadlineTask t = null;
```
Javier-Lua: Maybe use task instead of t for clarity? I noticed the same issue in several other places too.

src/main/java/clanker/Clanker.java
```diff
@@ -0,0 +1,283 @@
...
+        prompt.add("I found these tasks matching your search term!");
+
+        for (Pair<Integer, String> p : tasks) {
+            prompt.add(String.format("%s. %s", p.getKey() + 1, p.getValue()));
+        }
+
+        writePrompt(prompt.toArray(String[]::new));
+    }
+
+    private static void writePrompt(String... lines) {
```
Javier-Lua: writePrompt sounds like it makes changes. Maybe printPrompt or displayPrompt would be clearer?

src/main/java/clanker/Clanker.java
```diff
@@ -0,0 +1,283 @@
...
+        EventTask t = new EventTask(description, from, to);
+
+        todoList.addTask(t);
+
+        writePrompt("Added new task:",
+                t.toString(),
+                String.format("There are now %d tasks in your list.", todoList.size()));
+    }
+
+    private static void handleList() {
```
Javier-Lua: Maybe listTasks() would be more descriptive?

Javier-Lua (commented):
LGTM. Just some nits to fix.

src/main/java/clanker/Clanker.java
```diff
@@ -0,0 +1,283 @@
...
+        String description = String.join(" ", cmd.getAllParameters());
+
+        if (description.isEmpty()) {
+            writePrompt("Oops! The description of a task cannot be empty!");
+            return;
+        }
+
+        String deadline = cmd.getOptionValue("by");
+
+        if (deadline == null || deadline.isEmpty()) {
```
wuliyq, line 145: will it be sufficient to check deadline.isEmpty() only?

src/main/java/clanker/Clanker.java
wuliyq: agree :)

wuliyq (approved):
Very nice work! The OOP is very clear :)

src/main/java/clanker/Clanker.java
mattcce: Chose this method name primarily to keep the naming in convention with all other command handlers; I think that `listTasks` might break that expectation to a reader unnecessarily.

src/main/java/clanker/Clanker.java
mattcce: Unfortunately not, the return type of `getOptionValue` is nullable, and `isEmpty` is a `String` method (so it will fail if no `/by` option is supplied by the user).

Updated javadoc comments to reflect that.

src/main/java/fmt/CommandParser.java
mattcce: renamed to `parsers` for clarity 👍

PR #42, opened 2025-08-22
[haobuhaoo] iP

haobuhaoo:
# Lax 👍 
- **Lax**
- _Lax_
1. ~Lax~
2. <ins>lax</ins>
```ruby
Lax is great.
```
- [X] `Lax`
- [ ] [Link](https://google.com)
```ruby
    run();
}
```


data/data.txt
```diff
@@ -0,0 +1,6 @@
+todo | 1 | read book
+deadline | 0 | return book | 2025-06-06T12:00
+event | 1 | project meeting | 2025-06-06T00:00 | 2025-06-07T00:00
+deadline | 1 | assignment | 2025-03-03T13:00
+event | 0 | game match | 2025-08-02T13:30 | 2025-08-02T18:00
+deadline | 0 | watch video | 2025-08-24T23:59
```
joelchongg: Nice serialized format!

src/main/java/lax/Parser.java
joelchongg, line 1: why not put in command package?

src/main/java/lax/Storage.java
```diff
@@ -0,0 +1,123 @@
...
+        try (Scanner scanner = new Scanner(file)) {
+            int corrupted = 0;
+            while (scanner.hasNextLine()) {
+                String line = scanner.nextLine().trim();
+                if (line.isEmpty()) {
+                    continue;
+                }
+
+                String[] data = line.split("\\|");
+                boolean completed = data[1].trim().equals("1");
```
joelchongg: might have IndexOutOfBoundsException if the data is corrupted (no completion status). Might want to see if you should handle that?

src/main/java/lax/command/DeleteCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Prints the success message after an execution.
+     *
+     * @param t        The <code>Task</code> deleted.
+     * @param taskList The tasklist.
+     * @return A <code>String</code> message.
+     */
+    public String print(Task t, TaskList taskList) {
```
joelchongg: naming is abit off since its not really printing but returning a string to be printed, but that also means you have to change it everywhere which is troublesome ☹️

src/main/java/lax/task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return <li>"deadline | 1 | name | 2025-08-26T13:24" if completed</li>
+     *         <li>"deadline | 0 | name | 2025-08-26T13:24" if not completed.</li>
+     */
+    @Override
+    public String toFile() {
```
joelchongg: perhaps getSerializedFormat() is a better name?

joelchongg (commented):
LGTM! Just the naming of some functions doesn't really reflect what it does (but mainly due to the need to refactor from CLI to GUI application

src/main/java/lax/Parser.java
haobuhaoo: oh didnt think of that. since Parser class only purpose is to turn the string into a Command object, it should be in the command package which also reduces the number of import statements.

src/main/java/lax/command/LabelCommand.java
```diff
@@ -0,0 +1,62 @@
...
+public class LabelCommand extends Command {
+    /**
+     * The task number to be labeled.
+     */
+    private final String taskNumber;
+
+    /**
+     * The type of label. <code>true</code> if label is mark. <code>false</code> if label is unmark.
+     */
+    private final boolean mark;
```
isaacyclai: Perhaps marked would be a bit more appropriate?

isaacyclai (commented):
LGTM! Well done :)

PR #51, opened 2025-08-23
[xavychin] iP

xavychin:
# Leo

> "Success starts with a single step, and planning makes it count."

**Chat your way to organized days with Leo**:
- Clear and intuitive text-based interface
- Beginner-friendly
- ~easy~ _EFFORTELESS_ to use 

### Features

- [x] Task and event management
- [x] Task status updates
- [ ] Reminders
- [ ] Recurring tasks

### Setup

1. Download the application [here](https://github.com/xavychin/ip/releases/download/A-Jar/Leo.jar).
2. Double click on it.
3. Add your tasks.
4. Sit back and let Leo chat your way to a perfectly managed day! :thumbsup:

### Using the Code
The main program file is `Leo.java`
```java
public static void main(String[] args) throws IOException {
    new Leo(".data/Leo.txt").run();
}
```

### Testing the Code
There are various test cases that can be run in `src/Test/java/Leo` using either `gradle` or `IntelliJ`.

Additionally, `text-ui-test/runtest.bat` can be run to test the output of the application. Before using `text-ui-test/runtest.bat`, modify the parameter in `Leo.java`.
```java
public static void main(String[] args) throws IOException {
    new Leo("../.data/Leo.txt").run();
}
```


src/main/java/leo/functions/commands/AddTaskCommand.java
```diff
@@ -0,0 +1,106 @@
+package leo.functions.commands;
```
Shearer1509, line 1: good job seperating your import statements based on their package, which makes it easier to browse the list and determine the dependencies when there are many imports

src/main/java/leo/functions/search/Find.java
```diff
@@ -0,0 +1,54 @@
+package leo.functions.search;
+
+import leo.functions.task.TaskList;
+
+/**
+ * Represents the function that is called when the user prompts
+ * the chatbot to search the related tasks based on a keyword.
+ */
+public class Find {
```
Shearer1509: good job adding clear javaDoc comments that explains the functionality, parameters and return values of the code,

src/main/java/leo/display/UI.java
```diff
@@ -0,0 +1,52 @@
...
+
+        //Solution of using equalsIgnoreCase() suggested by IntelliJ code completion
+        String callFunction = "";
+        while (!callFunction.equalsIgnoreCase("bye")) {
+            callFunction = scanner.nextLine();
+            //Solution adapted from
+            // https://www.perplexity.ai/search/catch-a-function-but-handle-it-prjjRGnZRsu8igx_P1RE7A
+            try {
+                func.searchFunctions(callFunction);
+            } catch (ZeroLengthException
```
Shearer1509, line 40: i like your idea of capturing >= 1 exceptions while avoiding catch-them-all at the same time, this ensures that only expected expections are captured and its message parsed

src/main/java/leo/FileHandler.java
```diff
@@ -0,0 +1,179 @@
+package leo;
```
Shearer1509: perhaps this solution can be made simpler by first updating our TaskList, then using the TaskList as paramter in our save function to overwrite the existing data.txt each time, and load simply takes input from the data.txt and parses them to create a TaskList; however the effort put into this solution (which appends instead of overwrites) which is potentially more efficient is commendable

Shearer1509 (commented):
good job overall for this commit

src/main/java/leo/FileHandler.java
xavychin: Thank you for the suggestion 👍

src/Test/java/leo/DateTimeParserTest.java
```diff
@@ -0,0 +1,85 @@
+package leo;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.time.DateTimeException;
+
+import org.junit.jupiter.api.Test;
+
+public class DateTimeParserTest {
```
ryan-toh: Looks good! The naming conventions are mostly followed.

src/main/java/leo/functions/commands/AddTaskCommand.java
```diff
@@ -0,0 +1,204 @@
...
+     * Adds a ToDo task object to the list.
+     *
+     * @param userInput String containing the user input.
+     * @param listItems List of tasks.
+     * @throws ArrayIndexOutOfBoundsException If the user input is in the wrong format.
+     * @throws IOException If the file storing data cannot be found.
+     */
+    //Solution adapted from https://www.perplexity.ai/search/catch-a-function-but-handle-it-prjjRGnZRsu8igx_P1RE7A
+    public static void todo(String userInput, TaskList listItems) throws ArrayIndexOutOfBoundsException, IOException {
+        String[] userInputList = userInput.split("todo");
```
ryan-toh: The only difference between `todo()` and `todoReturnOutput()` seems to be that `todoReturnOutput()` returns a formatted string. If this is the case, I would suggest eliminating `todoReturnOutput()`, making `todo()` output a formatted string and the output could be ignored if the caller doesn't need it?

Disregard if there are other differences necessitating `todoReturnOutput()` though.

src/main/java/leo/functions/commands/Command.java
```diff
@@ -0,0 +1,9 @@
+package leo.functions.commands;
+
+/**
+ * The Command interface represents a contract for functions in the chatbot.
+ * Classes implementing this interface must provide mechanisms for either
+ * adding, deleting, marking or listing tasks.
+ */
+public interface Command {
+}
```
ryan-toh, line 9: Given that there are no common interface functions in `Command`, I would suggest not having an interface entirely to handle type of command, but instead to handle different task types? 

For example:
```
public interface Command {
}

public interface TaskCommand implements Command {
 String addTask(Task t) {}
 String deleteTask(Task t) {}
}

// ... continue with concrete classes for ToDo, Event, Deadline

public class ListCommand implements Command {
 String listAllTasks() {}
}
```

ryan-toh (commented):
Organised code! Suggested some changes to simplify the `Command` classes.

PR #53, opened 2025-08-23
[Peh Hou Jin] iP

203ZK:
### Travis (Requirement 1)

> "Life is like a box of chocolates; you never know what you're gonna get." - _Forrest Gump_ (Requirement 7, 10)

Here's some things you can do with [Travis](https://github.com/203ZK/ip) (Requirement 2, 8):
- Add/Delete tasks
- List existing tasks
- Have fun talking to a non-human!

Here's a sample code block (Requirement 4):

```java
public class Travis {
    private String name;

    public Travis(String s) {
        this.name = s;
    }

    @Override
    public String toString() {
        return "Hi, I'm " + this.name + "!";
    }
}
```

To set him up (Requirement 3):
1. Clone the repo
2. Make sure you have the necessary packages and JDK
3. Go wild~~dddd~~ 🙈 (Requirement 6)

Try modelling the following `TaskList` (Requirement 5, 9):
- [x] Buy bread
- [ ] Submit homework by 2 Feb 2024


src/main/java/travis/chatbot/Parser.java
```diff
@@ -0,0 +1,96 @@
...
+
+    // Patterns to match task formats
+    private static final Pattern TO_DO_PATTERN = Pattern.compile(RegexConstants.TO_DO_REGEX);
+    private static final Pattern DEADLINE_PATTERN = Pattern.compile(RegexConstants.DEADLINE_REGEX);
+    private static final Pattern EVENT_PATTERN = Pattern.compile(RegexConstants.EVENT_REGEX);
+
+    /**
+     * Parses the user's commands.
+     * If an existing command is not found, it treats the input as a new task to be added.
+     */
```
yoyongxuan, line 37: ### Should header comments have more details?

You may consider using @param, @return and @throws in this and other header comments

This is the format recommended by the [Java coding standard](https://se-education.org/guides/conventions/java/intermediate.html):

```
/**
 * Returns lateral location of the specified position.
 * If the position is unset, NaN is returned.
 *
 * @param x X coordinate of position.
 * @param y Y coordinate of position.
 * @param zone Zone of position.
 * @return Lateral location.
 * @throws IllegalArgumentException If zone is <= 0.
 */
public double computeLocation(double x, double y, int zone)
        throws IllegalArgumentException {
    // ...
}
```

src/main/java/travis/chatbot/Ui.java
```diff
@@ -0,0 +1,87 @@
...
+                this.warnMessage(e.getMessage());
+            } finally {
+                input = SCANNER.nextLine().trim();
+            }
+        }
+        this.SCANNER.close();
+        this.farewell();
+    }
+
+    // ------------------- WARNINGS -------------------
```
yoyongxuan, line 51: I like that you have added comments such as this to group certain methods together. It makes your code much easier to read.

src/main/java/travis/chatbot/Ui.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void notifyAddTask(String taskName, int numOfTasks) {
+        wrap(String.format(UiConstants.NEW_TASK, taskName) +
+                String.format(UiConstants.TOTAL_TASKS, numOfTasks));
+    }
+
+    public void notifyDeleteTask(String taskName, int numOfTasks) {
+        wrap(String.format(UiConstants.DELETED_TASK, taskName) +
+                String.format(UiConstants.TOTAL_TASKS, numOfTasks));
```
yoyongxuan: Should your line break be before your + operator?
According to Java Coding Guide, line breaks should be placed before any operators such as +.

Consider wrapping your code like this instead:
```suggestion
        wrap(String.format(UiConstants.NEW_TASK, taskName)
                + String.format(UiConstants.TOTAL_TASKS, numOfTasks));
    }

    public void notifyDeleteTask(String taskName, int numOfTasks) {
        wrap(String.format(UiConstants.DELETED_TASK, taskName)
                + String.format(UiConstants.TOTAL_TASKS, numOfTasks));
```

src/main/java/travis/chatbot/Parser.java
```diff
@@ -0,0 +1,96 @@
...
+    // Patterns to match user commands
+    private static final Pattern MARK_AS_DONE_PATTERN = Pattern.compile(RegexConstants.MARK_AS_DONE_REGEX);
+    private static final Pattern MARK_AS_NOT_DONE_PATTERN = Pattern.compile(RegexConstants.MARK_AS_NOT_DONE_REGEX);
+    private static final Pattern DELETE_TASK_PATTERN = Pattern.compile(RegexConstants.DELETE_TASK_REGEX);
+    private static final Pattern FIND_TASK_PATTERN = Pattern.compile(RegexConstants.FIND_TASK_REGEX);
+
+    // Patterns to match task formats
+    private static final Pattern TO_DO_PATTERN = Pattern.compile(RegexConstants.TO_DO_REGEX);
+    private static final Pattern DEADLINE_PATTERN = Pattern.compile(RegexConstants.DEADLINE_REGEX);
+    private static final Pattern EVENT_PATTERN = Pattern.compile(RegexConstants.EVENT_REGEX);
```
yoyongxuan: ### Should constants have a common prefix?

The Java Coding Standard states that
>  Associated constants should have a common prefix.

while your constants have a common suffix (PATTERN) instead.
Consider moving the word PATTERN to the front of each name to make it a prefix instead.

src/main/java/travis/constants/RegexConstants.java
```diff
@@ -0,0 +1,11 @@
+package travis.constants;
+
+public final class RegexConstants {
+    public static final String TO_DO_REGEX = "^todo (.*)$";
+    public static final String DEADLINE_REGEX = "^deadline (.*) /by (.+)$";
+    public static final String EVENT_REGEX = "event (.*) /from (.+) /to (.+)";
+    public static final String MARK_AS_DONE_REGEX = "^mark (\\d+)$";
+    public static final String MARK_AS_NOT_DONE_REGEX = "^unmark (\\d+)$";
+    public static final String DELETE_TASK_REGEX = "^delete (\\d+)$";
+    public static final String FIND_TASK_REGEX = "^find (.*)$";
```
yoyongxuan: ### Should constants have a common prefix?

Consider making your constants share a prefix instead of a suffix as mentioned above. 
There are a few other instances of code with a similar issue you may wish to amend.

src/main/java/travis/chatbot/Ui.java
```diff
@@ -0,0 +1,87 @@
...
+import travis.exceptions.InvalidTaskException;
+
+import java.util.Scanner;
+
+/**
+ * Represents the user interface of the chatbot.
+ * Users interact with the chatbot through CLI commands via this class.
+ */
+public class Ui {
+    private final Scanner SCANNER = new Scanner(System.in);
```
yoyongxuan: ### Did you mean for SCANNER to be considered a constant?

I noticed you have named SCANNER in SCREAMING_SNAKE_CASE as suggested by the [Java Coding Standard](https://se-education.org/guides/conventions/java/intermediate.html) for names of constants.

However, according to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names):

> Constants are static final fields whose contents are deeply immutable and whose **methods have no detectable side effects**. Examples include primitives, strings, immutable value classes, and anything set to null. If **any of the instance's observable state can change**, it is not a constant. Merely intending to never mutate the object is not enough.

Since the method Scanner.nextLine() has the side effect of prompting the user for an input and the String it returns depends on the contents of the user's input, it should not be considered a constant despite being final.

Consider using the camelCase naming style instead.

yoyongxuan (commented):
Overall, I found your code very neat, structured and easy to read. 
Other than a lack of header comments, I've only found a few minor violations of the Java Coding Standard.
Great job!

src/main/java/travis/constants/UiConstants.java
```diff
@@ -0,0 +1,14 @@
+package travis.constants;
+
+public final class UiConstants {
```
xavychin, line 3: I like how you group the constants used into a class. It makes your code cleaner and improves readability.

src/main/java/travis/tasks/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package travis.tasks;
```
xavychin, line 1: Good job on seperating your code into various packages making it easier to understand which part of the program the file is handling.

src/main/java/travis/tasks/Event.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public String getFileString() {
+        return String.join(
+                " | ", "E", this.getStatusIcon(), this.description, this.start, this.end) + "\n";
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (from: " + this.start + " to: " + this.end + ")";
```
xavychin, line 20: Nice use of `super.toString()` to keep the same output for all the subclasses of Task, reducing mistakes due to typing the same thing multiple times for the subclasses.

src/main/java/travis/storage/Storage.java
```diff
@@ -0,0 +1,100 @@
...
+                    yield toDoLoader.load(line);
+                }
+                case 'D' -> {
+                    DeadlineLoader deadlineLoader = new DeadlineLoader();
+                    yield deadlineLoader.load(line);
+                }
+                default -> {
+                    EventLoader eventLoader = new EventLoader();
+                    yield eventLoader.load(line);
+                }
```
xavychin: Perhaps you could create a new case for `EventLoader` instead of using default?

xavychin (commented):
Overall good job for the commit 👍

PR #64, opened 2025-08-23
[Glenn-Chiang] iP

Glenn-Chiang:
# Glendon Bot
> "This is the greatest application of all time." -- Alan Turing

 ⚡**Blazingly fast** CLI-based task manager chatbot for those who thrive in the terminal.

## Features
- Add tasks (todos, deadlines, events)
- List tasks
- Delete tasks
- Mark and unmark tasks
- Save and load tasks to and from file
- Search for tasks  

### Future enhancements:
- [ ] GUI
- [ ] Use varargs

## Installation
Install jar file from [here](https://github.com/Glenn-Chiang/ip/releases/tag/Level-9)

## Local Usage
1. Clone the repository
2. Run `./gradlew run` in the terminal at the project root

Main function here for no particular reason:
```java
public static void main(String[] args) throws GlendonException {
        new Glendon(tasksDataPath).run();
}
```


src/main/java/glendon/Storage.java
```diff
@@ -0,0 +1,171 @@
...
+            break;
+        case 4:
+            // Deadline
+            String dateStr = components[3];
+            LocalDate date = LocalDate.parse(dateStr, dateFormat);
+            task = new Deadline(description, date);
+            break;
+        case 5:
+            // Event
+            String startStr = components[3];
```
KH-boop-bit, line 82: Perhaps a more intuitive naming convention here, instead of startStr, could be startDateString to specify that this string is indeed a date string

src/main/java/glendon/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+        return Arrays.stream(Glendon.Command.values())
+                .filter(cmd -> cmd.keyword.equals(commandKeyword))
+                .findFirst()
+                .orElse(null);
+    }
+
+    /**
+     * Extracts and returns the specified index from the input string.
+     * Commands MARK, UNMARK and DELETE require an index to be specified as the second argument in the input string.
+     */
```
KH-boop-bit, line 38: As per the header comments below, do specify what @param and @return is in its respective segments

src/main/java/glendon/task/Task.java
```diff
@@ -0,0 +1,37 @@
+package glendon.task;
+
```
KH-boop-bit, line 2: Header comments could be useful here to denote what each method does, seeing as they are not present in the respective Task subclasses as well

src/main/java/glendon/task/Event.java
```diff
@@ -0,0 +1,41 @@
...
+import glendon.Storage;
+
+public class Event extends Task {
+    private static final DateTimeFormatter dateTimeFormat = DateTimeFormatter.ofPattern("MMM d yyyy HHmm");
+    private final LocalDateTime start;
+    private final LocalDateTime end;
+
+    public Event(String description, LocalDateTime start, LocalDateTime end) throws GlendonException {
+        super(description);
+        this.start = start;
```
KH-boop-bit, line 19: A better naming convention for start could be startDate to further define the parameter

src/test/java/glendon/ParserTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+public class ParserTest {
+    @Test
+    public void parseTask_validTodo_success() throws Exception {
+        assertEquals("[T][ ] buy new keyboard", Parser.parseTask("todo buy new keyboard").toString());
```
KH-boop-bit, line 19: would be good to break after operators, like the "," to improve readability

src/main/java/glendon/Glendon.java
```diff
@@ -0,0 +1,172 @@
+package glendon;
+
+import java.util.List;
+import java.util.Scanner;
+
+import glendon.task.Task;
+
+public class Glendon {
```
KH-boop-bit, line 11: This file's styling and naming conventions are clean!

src/main/java/glendon/Glendon.java
```diff
@@ -0,0 +1,172 @@
...
+        while (true) {
+            String input = scanner.nextLine();
+            Command command = Parser.parseCommand(input);
+
+            if (command == null) {
+                Ui.displayUnknown();
+                continue;
+            }
+
+            try {
```
NihaalManaf: Really appreciate the super clean code here 👍

src/main/java/glendon/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+import glendon.task.ToDo;
+
+public class Parser {
+    private static final DateTimeFormatter dateFormat = DateTimeFormatter.ISO_LOCAL_DATE;
+    private static final DateTimeFormatter dateTimeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+
+    /**
+     * Extracts and returns the type of command from the input string.
+     */
+    public static Glendon.Command parseCommand(String input) {
```
NihaalManaf, line 27: Including what the @param and @return type here would have been helpful. Noticed this issue across a few other methods (but def not all)

src/main/java/glendon/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+        String commandKeyword = input.split(" ")[0];
+        return Arrays.stream(Glendon.Command.values())
```
NihaalManaf, line 29: Oooh interesting way to implement this haha. I've unfortunately got skill issues.. :)

src/main/java/glendon/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+                throw new GlendonException("Invalid event start or end datetime format");
+            }
+            break;
+        default:
+            throw new GlendonException("Invalid task format");
+        }
+        return task;
+    }
+
+    private static boolean validateInput(Pattern pattern, String input) {
```
NihaalManaf, line 118: With regards to the coding standard, perhaps the names sound more like boolean names/methods by phrasing like a question? (e.g isTodoInputValid)? Makes it easier to understand it too!

This applies for the next couple of methods return bool values

src/main/java/glendon/Storage.java
NihaalManaf: Yeah I agree!

src/main/java/glendon/Ui.java
```diff
@@ -0,0 +1,76 @@
+package glendon;
+
+import java.util.List;
+
+import glendon.task.Task;
+
+public class Ui {
+    public static void intro(String name) {
```
NihaalManaf: I believe for methods that does something, there ought to be a verb accompanying the task. So in this case, it should be, 

`public static void sendIntro(String name)`

Comment applies or a few more of the methods in this class

src/main/java/glendon/Glendon.java
```diff
@@ -0,0 +1,172 @@
...
+            this.taskList = new TaskList();
+        }
+    }
+
+    /**
+     * Reads user commands and executes responses.
+     *
+     * @throws GlendonException
+     */
+    public void run() throws GlendonException {
```
NihaalManaf: should GlendonException be more named in a more specific way? Unless this was implied to catch any exception thrown by the whole Glendon chatbot? I would say this may not also be recommended. (not really a coding standard violation though)

> Do not to catch them all - Prof Boyd

src/test/java/glendon/GlendonTest.java
```diff
@@ -0,0 +1,12 @@
+package glendon;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class GlendonTest {
+    @Test
+    public void dummyTest() {
```
NihaalManaf: I don't think you meant to push this haha

NihaalManaf (commented):
Overall, most coding standards have been followed! Cool. Few changes here and there and it'd be perfect.

PR #70, opened 2025-08-23
[gab-er] iP

gab-er:
# Jimmy 
> The chatbot for your task scheduling needs 

Jimmy is a chatbot to handle your task scheduling needs! It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use 🔥 

All you need to do is,
1. download it from [here](https://github.com/gab-er/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😄 

Best of all, its **FREE!**

Features:
- [ ] Managing tasks 
- [ ] Managing deadlines
- [ ] Reminders (coming soon) 

#
```java
public class Main {
       new Jimmy().run();
    }
}
```


src/main/java/jimmy/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Represents a Storage object.
+ */
+public class Storage {
+    private final File storageFile;
+
+    /**
+     * Constructor for a Storage object.
+     *
```
AaronToh, line 29: Any reason you began your comment with "Constructor" instead of "Constructs"?

src/main/java/jimmy/task/Deadline.java
```diff
@@ -0,0 +1,61 @@
...
+    private LocalDateTime deadline;
+
+    /**
+     * Constructor for a Deadline object.
+     *
+     * @param description The description of the task.
+     * @param deadline    The deadline of the task in the format yyyy-mm-dd.
+     */
+    public Deadline(String description, boolean completed, String deadline) throws JimmyException {
+        super(description, completed);
```
AaronToh: Perhaps you could rename "completed" as "isComplete"

src/main/java/jimmy/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param taskList TaskList of stored tasks.
+     */
+    public void saveData(TaskList taskList) {
+        try {
+            FileWriter fw = new FileWriter(this.storageFile);
+            for (int i = 0; i < taskList.size(); i++) {
+                fw.write(taskList.getTask(i + 1).toStorageString()); // Add 1 to account for zero-indexed position
+                fw.write(System.lineSeparator()); // Start next task on a new line
+            }
```
AaronToh, line 176: I like how you wrote comments to improve readability

src/main/java/jimmy/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+            // Update the arraylist to store the tasks
+            while (s.hasNext()) {
+                Task newTask = readData(s.nextLine());
+                taskList.addTask(newTask);
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("File not found");
+        } catch (JimmyException e) {
+            System.out.println(e);
+        }
```
AaronToh, line 70: Any reason you chose to catch errors in your Storage file instead of your Jimmy file?

AaronToh (commented):
LGTM. You could look into creating a command class to separate some task logic from your parser.

src/main/java/jimmy/Storage.java
gab-er: Oops, thanks for letting me know, will update!

src/main/java/jimmy/task/Deadline.java
gab-er: Yep, thanks for pointing that out!

src/main/java/jimmy/Storage.java
gab-er: I thought that it would be best to have the errors that occur from methods related to the Storage class to be caught in the Storage class itself, but I will take note of this in the future, thanks!

PR #86, opened 2025-08-23
[joelchongg] iP

joelchongg:
# ChatterBox
> “Ideas belong in your head, tasks belong here.”
Chatterbox helps you remember what you need, ensuring you don't forget anything. It's
- Easy to use 😃 
- Easy to learn 📖 
- Easy to rely on 💯 

All you need to do is
1. Install it from [here](https://github.com/joelchongg/ip/releases/tag/A-Jar)
2. run the software
3. add your tasks
4. let it manage your tasks for you!

Best part, it's **FREE** and *super simple*!  
(~~No hidden costs~~ ✅)

Features:
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Storage of tasks for future use

Below is the `main` method for running the program:
```java
public class ChatterBox {
        initialize();
        ChatterBoxUI.greet();
        run();
        ChatterBoxUI.farewell();
    }
}
```


src/main/java/chatterbox/memory/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public T removeItem(int index) throws IndexOutOfBoundsException {
+        return this.storage.remove(index);
+    }
+
+    public T getItem(int index) throws IndexOutOfBoundsException {
+        return this.storage.get(index);
+    }
+
+    public int size() {
```
t-yunqi, line 75: Should this be renamed getSize() as it's a method?

src/main/java/chatterbox/command/CommandProcessor.java
```diff
@@ -0,0 +1,372 @@
...
+                    "Uh oh! You forgot to include a description for your event task! Try again!"
+                );
+            }
+
+            String[] tokens = parseInput(input, " /from ", " /to ");
+
+            if (tokens.length != 3) {
+                throw new ChatterBoxException(
+                    "Uh oh! You did not input your event task correctly!"
+                    + "Try: event <description> /from <time> /to <time>"
```
t-yunqi, line 257: Should the indentation for these lines be 8 spaces, since it's part of a wrapped line from parent line 249? Same comment for similar lines above this one, where argument is placed a line after the method name.

src/main/java/chatterbox/memory/MemoryStorage.java
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param index Index in which the task is stored in memory
+     * @param isCompleted Completion status to set the selected task to
+     */
+    public static void updateTaskCompletion(int index, boolean isCompleted) {
+        File tempFile = new File("./data/tasks_tmp.txt");
+        int currentLine = 0;
+
+        try (Scanner scanner = new Scanner(taskFile);
+             PrintWriter writer = new PrintWriter(new FileWriter(tempFile))) {
```
t-yunqi, line 93: Should try blocks be unparameterized according to the coding standard? (though I do prefer this since its safer!)
You can use a finally block instead to ensure these resources are closed at the end of the block. Same comment for other try-with-resources statements in this class.

src/main/java/chatterbox/memory/MemoryStorage.java
```diff
@@ -0,0 +1,241 @@
...
+     * Adds and stores a new task into memory.
+     * Warns the user via the command line interface if task cannot be saved.
+     *
+     * @param task Task to be saved into memory
+     */
+    public static void saveTask(Task task) {
+        try (PrintWriter writer = new PrintWriter(new FileWriter(taskFile, true))) {
+            String line = task.getTaskSymbol() + " | "
+                    + (task.getStatusIcon().equals("X") ? "1" : "0") + " | "
+                    + task.getTaskDescription();
```
t-yunqi, line 129: I like how you placed these line breaks!

t-yunqi (commented):
LGTM! Your code is very clean👍

PR #90, opened 2025-08-24
[niyniy123] iP

niyniy123:
# JibJab
> "In heaven, all the interesting people are missing."  &ndash; Friedrich Nietzsche

JibJab helps you keep track of tasks. It's
- CLI-based
- **easy** to learn

All you need to do is,

1. Download it from [here](https://google.com)
2. Run it
3. Save your tasks

Features:

- [x] Managing tasks
- [ ] Task suggestions (Coming soon)

To remove all tasks temporarily, `rm -rf /` should help

```java
public class Main {
        System.out.println("Hello World");
    }
}
```


src/main/java/jibjab/Deadline.java
```diff
@@ -0,0 +1,44 @@
...
+ * Represents a task with a specific deadline by which it must be completed.
+ * A Deadline extends the Task class and includes a due date and time
+ * to indicate when the task should be finished.
+ *
+ * @author niyniy123
+ */
+public class Deadline extends Task {
+
+    protected LocalDateTime by;
+
```
ziheng119, line 23: Perhaps you could name the variable name better?

src/main/java/jibjab/JibJab.java
```diff
@@ -0,0 +1,164 @@
...
+                break;
+            default:
+                throw new JibJabException("I don't understand this command");
+            }
+            return out.toString();
+        } catch (JibJabException e) {
+            return ui.showError(e.getMessage());
+        }
+    }
+
```
ziheng119: I like how you used switch instead of multiple if-else!! LGTM!

src/main/java/jibjab/Task.java
```diff
@@ -0,0 +1,61 @@
...
+ * Represents a task in the JibJab application.
+ * This is the base class for all task types (ToDo, Deadline, Event) and provides
+ * common functionality for task description and task completion.
+ *
+ * @author niyniy123
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
```
ziheng119, line 20: Should the access modifier be protected?

ziheng119 (commented):
Good job on the iP so far! LGTM!

src/main/java/jibjab/Deadline.java
niyniy123: Thanks! I have renamed it now

PR #91, opened 2025-08-24
[Sengernest] ip 

Sengernest:
# Sengernest

> “If you fail to plan, you plan to fail.” – Benjamin Franklin 

Sengernest frees your mind from having to remember things you need to do so that you can better plan ahead! It's:

- **Text-based**
- **Easy to learn**
- **~~FAST~~ SUPER FAST** to use 🚀

All you need to do is:

1. Download Sengernest [here](https://google.com)
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉

## Features
- [X] Managing tasks  
- [ ] Managing deadlines *(coming soon)*  
- [ ] Reminders *(coming soon)*  

> [!NOTE]
> If you are a Java programmer, you can use it to practice Java too.  
Here's the main method:

```java
public class Main {
    }
}

To run the program, type `java Main` in your terminal.

### CODE REVIEW

File: src/main/java/sengernest/commands/UnmarkCommand.java [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+     * Constructs an UnmarkCommand from a string representing the task number.
+     *
+     * @param indexStr The 1-based task number provided by the user.
+     * @throws IllegalArgumentException if the provided string is not a valid number.
+     */
+    public UnmarkCommand(String indexStr) {
+        try {
+            this.index = Integer.parseInt(indexStr) - 1;
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Enter a valid number after 'mark'.");
```
junhao4, line 29: Should this be "Enter a valid number after 'unmark'." ?

src/main/java/sengernest/Sengernest.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param input The user input string.
+     * @return The response message from the bot, or an error message if parsing or execution fails.
+     */
+    public String getResponse(String input) {
+        try {
+            Command command = Parser.parse(input);
+            command.execute(tasks, ui, storage);
+            return ui.getLastMessage();
+        } catch (Exception e) {
```
junhao4, line 46: I noticed you catch all exceptions in many places which can be potentially dangerous. Perhaps you can create a parent exception that all your exceptions extend from and catch that instead.

src/main/java/sengernest/tasks/Task.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Represents a generic task.
+ */
+public class Task {
+    /** The description of the task. */
+    private final String tasking;
+
+    /** Whether the task is finished. */
+    private boolean finished;
```
junhao4: Should this boolean be isFinished instead

src/main/java/sengernest/tasks/Task.java
```diff
@@ -0,0 +1,89 @@
...
+    public boolean isEmpty() {
+        return this.tasking.isEmpty();
+    }
+
+    /**
+     * Checks if the task has been marked as finished.
+     *
+     * @return true if the task is finished, false otherwise.
+     */
+    public boolean isFinished() {
```
junhao4, line 37: then this could be getIsFinished?

src/main/java/sengernest/Sengernest.java
```diff
@@ -0,0 +1,48 @@
+package sengernest;
```
junhao4, line 1: Overall well written code, it was very hard to find anything.

PR #117, opened 2025-08-24
[junhao4] iP

junhao4:
# Mark
> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

Mark frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~FAST~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/nus-cs2103-AY2526S1/ip/pull/117).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [X] Managing tasks
- [ ] Reminders (coming soon)
---
```
public class Main {
    }
}
```


src/main/java/command/Command.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public abstract class Command {
+    String arg;
+    TaskList taskList;
+
+    /**
+     * Creates a new <code>Command</code> with the given arguments and task list.
+     *
+     * @param arg      The user input argument for the command
+     * @param tasklist The <code>TaskList</code> that this command will operate on
```
flyingcat9, line 21: I just noticed that throughout the code, some nouns are surrounded by the <></> notation in the comments and wondered if it was intentional?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,102 @@
+package storage;
+
+import exceptions.FileLoadException;
+import task.Task;
+import task.TaskList;
+import task.TodoTask;
+import task.DeadlineTask;
+import task.EventTask;
+
+import java.io.*;
```
flyingcat9: Perhaps you could be more specific as to which class you imported from the java.io library?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,81 @@
...
+        );
+    }
+
+    /**
+     * Prints a message with surrounding lines.
+     *
+     * @param msg The message to print
+     */
+    public static void print(String msg) {
+        System.out.println(
```
flyingcat9: I am guessing the spacing is for consistency with the other methods. Just, as a reader, I feel like the spacing could perhaps be further enhanced upon.

src/test/java/mark/MarkTest.java
```diff
@@ -0,0 +1,17 @@
+package mark;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MarkTest {
+    @Test
+    public void dummyTest() {
```
flyingcat9: I think you did a great job with most of the code. Perhaps the test method name could be more descriptive?

src/main/java/exceptions/EmptyDescriptionException.java
```diff
@@ -0,0 +1,8 @@
+package exceptions;
```
flyingcat9, line 1: I like about how you have so many classes for the exceptions so that users know how to change their command.

flyingcat9 (commented):
I think your code was readable and easy to understand. There are just a few minor things here and there but overall, well done.

PR #120, opened 2025-08-24
c-j-lh iP

chleheng:
# Rama2
> “Fast like a butterfly” 

Rama2 frees your mind of having to remember things you need to do. It's,

- CLI
- cute
- simple

All you need to do is,

1. download it from [here](https://github.com/c-j-lh/ip/)  <todo>.
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😀

And it is **FREE**!

Features:

- [x]  Managing tasks
- [ ] Managing deadlines
- [ ] Reminders (coming soon)
- [ ] If you are a Java programmer, you can use it to practice Java too. Here's the `main` method:

```
public class Main {
    }
}
```


src/main/java/ramarama/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.time.LocalDate;
+
+/*
+ * Parses  single raw input String into Cmd (with a,b,c fields).
+ */
+class Parser {
+    static class Cmd {
+        String name; // list, bye, todo, deadline, event, mark, unmark, delete, unknown
+        String a;
```
JunhaoW3: Perhaps this can be changed to "String description" to better describe the value you are trying to store in "a"

src/main/java/ramarama/Rama2.java
```diff
@@ -0,0 +1,171 @@
...
+    /*
+     * Main runner function
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean exit = false;
+        while (!exit) {
+            String input = ui.read();
+            ui.line();
+            Parser.Cmd c = Parser.parse(input);
```
JunhaoW3: Maybe it is possible to account for different capitalization of String input using .toLowerCase()

src/main/java/ramarama/Rama2.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    private void storageSave() {
+        try {
+            storage.save(tasks);
+        } catch (Exception e) {
+            /* ignore minimal */
+        }
+    }
```
JunhaoW3, line 188: Should this be extracted out to Storage?

src/main/java/ramarama/Task.java
```diff
@@ -0,0 +1,27 @@
...
+import java.time.LocalDate;
+
+/*
+ * Task, with type, markedness, description, and either date-as-string or date-as LocalDate
+ */
+class Task {
+    // char type; // 'T','D','E'
+    enum TaskType {
+        T, D, E
+    };
```
JunhaoW3: Good use of enumeration here

src/main/java/ramarama/Ui.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
+
+/*
+ * Stores various standard output strings.
+ */
+class Ui {
+    final Scanner in = new Scanner(System.in);
+    final String m = "    ____________________________________________________________";
```
JunhaoW3, line 10: Maybe instead of using "String m", you can use a variable name that is more easily understand, like "printHorizontalLine"

JunhaoW3 (commented):
Looks good, only some simple changes to be made

PR #128, opened 2025-08-24
[Jarell] iP

liawjarell:
# Mochi
Your all in one chatbot that helps you keep track of your tasks!
> MochiMochiMochi

- Command-line-interface application
- Easy to learn
- Supper 

All you need to do is:
1. download it from [here}(https://github.com/liawjarell/ip/releases/tag/Level-9)
2. double click it
3. add ur tasks
4. let it manage your tasks for u! 🔥 

And it is **FREE**!

Mochi supports the following tasks:
- **todo** task `todo read book`
- **deadline** task  `deadline return book /by 2025/11/26 1800
- **event task** `event attend class /from 2pm /to 4pm

Mochi also supports the following operations:
- [x] exit program `bye`
- [x] `list` shows all existing tasks
- [x] `mark 1` marks the task number 1 as done
- [x] `unmark 1` unmarks the task number 1 as done
- [x] `delete 1` deletes the task number 1
- [ ] `find book` finds all tasks with "book" in the description 
- [ ] graphical user interface

This is the main method:
```ruby
    new Mochi("data/tasks.txt").run();
}
```


src/main/java/mochi/Mochi.java
```diff
@@ -0,0 +1,235 @@
...
+        this.storage.saveTasks(this.tasks.getTasks());
+    }
+
+    /**
+     * Loop that processes user input and invokes the appropriate methods to handle the input.
+     * This method delegates the task of parsing user input to the Parser class.
+     * It continuously reads user input until the user decides to exit the program.
+     */
+    private void getInput() {
+        // Variables
```
JothamWong: This comment style of declaring // Variables does not seem to be followed across the codebase. I feel it should either be consistently practiced or removed.

src/main/java/mochi/Mochi.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * The entry point for the Mochi task management application.
+     * This method initializes the application with a specified task storage file
+     * and starts the main program logic by invoking the {@code run()} method.
+     *
+     * @param args Command-line arguments passed to the program.
+     *             Currently not used by the Mochi application.
+     */
+    public static void main(String[] args) {
```
JothamWong: Unused methods should not be present in the code. With git, we can delete this function and easily restore it when we do eventually need it

src/main/java/mochi/gui/MainWindow.java
```diff
@@ -0,0 +1,65 @@
...
+                DialogBox.getUserDialog(input, userImage),
+                DialogBox.getDukeDialog(response, mochiImage)
+        );
+
+        if (input.trim().equals("bye")) {
+            javafx.animation.PauseTransition delay =
+                    new javafx.animation.PauseTransition(javafx.util.Duration.seconds(1));
+            delay.setOnFinished(e -> javafx.application.Platform.exit());
+            delay.play();
+        }
```
JothamWong, line 59: I really like this! It's a fun animation!

src/main/java/mochi/parser/Parser.java
```diff
@@ -0,0 +1,269 @@
...
+            return mochi.printList();
+        } else {
+            // Takes care of white spaces before description
+            String command = input.split("\s+", 2)[0];
+
+            switch (command) {
+            case "mark", "unmark", "delete":
+                return parseNumberedAction(mochi, input, command);
+            case "todo", "deadline", "event":
+                return parseAddTask(mochi, input, command);
```
JothamWong: This is an interesting way of handling the common task parsing in one place, but I feel that it causes too much nesting.
I would personally have a flat switch case instead of a nested switch case which you have in the parseAddTask function for example.

src/main/java/mochi/parser/Parser.java
```diff
@@ -0,0 +1,269 @@
...
+            if (dateString.contains("/")) {
+                date = LocalDate.parse(dateString, DateTimeFormatter.ofPattern("yyyy/MM/dd"));
+            } else if (dateString.contains("-")) {
+                date = LocalDate.parse(dateString);
+            } else if (dateString.equals("tomorrow")) {
+                date = LocalDate.now().plusDays(1);
+            } else if (dateString.equals("today")) {
+                date = LocalDate.now();
+            } else {
+                throw new MochiException("Please input a proper date and time in the format [yyyy/MM/dd HHmm]");
```
JothamWong, line 294: It's really cool that you handled the different keywords like today or tomorrow. I did not consider doing this

src/main/java/mochi/task/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+ * This class is responsible for storing and retrieving tasks.
+ * It also provides methods for manipulating the task list, such as adding, deleting,
+ * marking, and unmarking tasks.
+ */
+public class TaskList {
+
+    /**
+     * A List object that represents the list of tasks currently managed by Mochi.
+     */
+    protected List<Task> tasks;
```
JothamWong, line 20: Nice use of List over ArrayList

src/main/java/mochi/task/ToDos.java
```diff
@@ -0,0 +1,63 @@
+package mochi.task;
+
+import mochi.exception.MochiException;
+
+/**
+ * ToDos class for storing and retrieving tasks.
+ */
+public class ToDos extends Task {
```
JothamWong: I don't think this is great naming, because it implies that this class has multiple nouns, when it really only represents a single ToDo

src/main/java/mochi/ui/Ui.java
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @param message The message to be wrapped.
+     * @return The wrapped message.
+     */
+    public static String wrap(String message) {
+        return String.format("""
+                ________________________________________________________________________________
+                %s
+                ________________________________________________________________________________""",
+                message).indent(4);
```
JothamWong, line 30: I did not know about the indent method

src/main/java/mochi/ui/Ui.java
```diff
@@ -0,0 +1,175 @@
...
+    public void warn(String message) {
+        this.wrapPrint(message);
+    }
+
+    /**
+     * Prints a welcome message to the user.
+     */
+    public String welcome() {
+        return Messages.MESSAGE_WELCOME;
+        //        this.wrapPrint(Messages.MESSAGE_WELCOME);
```
JothamWong: With git, one should delete lines of code instead of have commented out lines, generally speaking.

src/main/java/mochi/ui/Ui.java
```diff
@@ -0,0 +1,175 @@
...
+        if (tasks.isEmpty()) {
+            return Messages.MESSAGE_EMPTY_LIST;
+            //            this.wrapPrint(Messages.MESSAGE_EMPTY_LIST);
+        } else {
+            return String.format(Messages.MESSAGE_LIST_PRINT, tasks.toString());
+            //            this.wrapPrint(String.format(Messages.MESSAGE_LIST_PRINT, tasks.toString()));
+        }
+    }
+
+    //    public static void main(String[] args) {
```
JothamWong: This should be deleted instead of commented out, since we have git

src/test/java/mochi/parser/ParserTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    public void parseGeneralInput_numberedTask_invalidIndexExceptionThrown() {
+        StubMochi mochi = new StubMochi();
+        mochi.seedTasks(2);
+
+        assertThrows(MochiException.class, () -> Parser.parseGeneralInput(mochi, "mark 3"));
+        assertThrows(MochiException.class, () -> Parser.parseGeneralInput(mochi, "unmark 3"));
+        assertThrows(MochiException.class, () -> Parser.parseGeneralInput(mochi, "delete 3"));
+        assertThrows(MochiException.class, () -> Parser.parseGeneralInput(mochi, "mark 0"));
+    }
```
JothamWong, line 52: Really nice use of the stub to test for method calls

JothamWong (commented):
Overall, I found your code easy to read but I don't agree on the use of nested switch cases for the handling of the add task commands and so on. It forces the user to mentally jump through one more function than necessary. Otherwise, the code is good!
I noted a few minor coding standard violations too.

PR #133, opened 2025-08-25
[Junronggg] iP

Junronggg:
# Junny

> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://nus-cs2103-ay2526s1.github.io/website/schedule/week4/project.html))

Junny frees your mind from having to remember things you need to do. It's,
- text-based  
- easy to use  
- ~~FAST~~ _SUPER FAST_ to use 

All you need to do is:
1. Download it from [here](https://github.com/Junronggg/ip)  
2. Double-click it  
3. Add your tasks  
4. Sit back and let it manage your life 😉

And it is **FREE**!

## Features
- [x] Managing tasks  
- [ ] Reminders (coming soon)  

---

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Junny/Junny.java [hunk 1]
```diff
@@ -0,0 +1,228 @@
...
+
+public class Junny {
+    // make it global, so all mthds can access and change
+    // the fields need to be static, so can be accessed in the main mthd
+
+    // when interact with Junny.Junny, each time there is a 'interactive board', which is the object of Junny.Ui class
+    static Ui ui = new Ui();
+    // for parsing commands
+    // static Junny.Parser parser = new Junny.Parser(ui);
+    // for file storage
```
josephkwok001: Perhaps consider declaring its accessibility, whether public or private fields

src/main/java/Junny/Junny.java
```diff
@@ -0,0 +1,228 @@
...
+                // handle exception 2
+                ui.printError("I'm sorry, but I don't know what that means :(");
+            }
+        }
+    }
+
+    public static void printLine() {
+        System.out.println("____________________________________________________________");
+    }
+
```
josephkwok001: You can possibly create a new TaskList class that handles all the execution of your tasks, such as markDone, markUndone, addTask, etc. That way, you can directly call your tasklist.function_name(), allowing for greater abstraction.

src/main/java/Junny/Junny.java
```diff
@@ -0,0 +1,228 @@
...
+            } catch (IllegalArgumentException e) {
+                // handle exception 2
+                ui.printError("I'm sorry, but I don't know what that means :(");
+                continue;
+            }
+
+            if (command == CommandTypes.BYE) {
+                ui.printBye();
+                break;
+            } else if (command == CommandTypes.LIST) {
```
josephkwok001: it seems you use a lot of if statements for different cases, perhaps you can use the Java enum to handle different cases.

src/main/java/Junny/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+            // handle exception 2
+            ui.printError("I'm sorry, but I don't know what that means :(");
+            return null;
+        } catch (Exception e) {
+            ui.printError("I'm sorry, but I don't know what that means :(");
+            return null;
+        }
+
+        switch (command) {
+            case CommandTypes.BYE:
```
josephkwok001: violate coding standard

src/main/java/Junny/Storage.java
josephkwok001, line 1: Good use of abstraction, easy to understand and simple

PR #144, opened 2025-08-25
[jovnc] iP

jovnc:
# Hermione

> "Books and cleverness! There are more important things - friendship and bravery." - Hermione Granger ([source](https://bookroo.com/quotes/hermione-granger))

Hermione frees your mind of having to remember things you need to do. It's,
* text-based
* easy to learn
* ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://nus-cs2103-ay2526s1.github.io/website/schedule/week4/project.html).
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features: 

- [ ] Managing tasks
- [ ] Reminders (coming soon)

---

```java
public class Hermione {
        startConsole(); // or startGui()
    }
}
```


src/main/java/hermione/validators/TaskValidator.java
adityagk753, line 1: I like how you put all the validation for the user inputs in a separate class, TaskValidator. This keeps the code really neat and modularised.

src/main/java/hermione/Hermione.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Main method to run the Hermione application.
+     * This method creates an instance of Hermione with the specified task storage file path
+     * and starts the application by calling the run method.
+     *
+     * @param args Command line arguments.
+     */
+    public static void main(String[] args) {
+        if (args.length > 1) {
+            System.out.println("Usage: java -jar hermione.jar [--console]");
```
adityagk753, line 37: I like how you provide the user with very descriptive feedback about the correct command structure. This makes the user's life much easier when they input a wrong command syntax.

src/main/java/hermione/validators/TaskValidator.java
adityagk753: Perhaps it may be a better design to put the validation inside the respective task's classes (as a static method)?

If you add new types of tasks (eg. another task which inherits from Task), you would currently need to add a new function to this TaskValidator class as well.

This would be better for modularisation in my opinion.

adityagk753 (commented):
Your code is really neat and compact. Especially love the level of modularisation and comprehensive user input validation you've done.

bluemayo (commented):
I like the meaningful and short names.

PR #148, opened 2025-08-25
[ Koh Wai Kei ] iP

yk-tuturu:
### MikuBot

> Miku miku oooeeeoooo~

Have the one and only Hatsune Miku as your personal task tracker companion. It's: 

- Very easy to use
- Very cute
- Very friendly 

All you need to do is: 

1. Download it from [here ](https://github.com/yk-tuturu/ip/tree/master)
2. Open the jar application
3. Start entering your commands and tasks 

Features: 
- [x] Managing tasks
- [x] Keeping track of dates
- [ ] Sending reminders

If you are a Java programmer, feel free to contribute
```
public class Main {
        System.out.println("Hello sekai! Miku is here!")
    }
}
```


src/main/java/miku/MikuBot.java
```diff
@@ -0,0 +1,113 @@
...
+            saveData.init();
+        } catch (FileIOError e) {
+            ui.print(e.getMessage());
+        }
+    }
+
+    /**
+     * Clears all saved tasks from the save file. Mostly for running tests.
+     */
+    public void ClearSave() {
```
plingplingshiny: Should method names be in camelCase? Maybe `clearSave()` might be better? Same for `RunCommand(String input)` and `Run()` too?

plingplingshiny (commented):
Nice job adding Javadoc comments to all the key public classes and even the `toString()` methods for your different Task types!
Just a small note: a few class methods aren’t in camelCase. Other than that, the code looks clean and well-structured 😃

src/main/java/miku/MikuBot.java
```diff
@@ -0,0 +1,113 @@
...
+        saveData.clearSave();
+    }
+
+    /**
+     * Executes a single command (used for unit testing or programmatic input).
+     *
+     * @param input the command string (required)
+     * @return the output message or error message
+     */
+    public String RunCommand(String input) {
```
ZayK1: Method names should follow camelCase convention. Should this be runCommand() instead of RunCommand()?

src/main/java/miku/MikuBot.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Runs the interactive console loop.
+     * <p>
+     * Prints the intro, populates tasks from save file, and repeatedly reads user input
+     * to execute commands until an exit command is received.
+     * </p>
+     */
+    public void Run() {
```
ZayK1: Should this be ```run()``` instead of ```Run()```?

src/main/java/miku/command/Command.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Alternate constructor
+     * @param key the keyword for the command
+     * @param usage the correct usage for the command, provided in errors and help functions
+     * @param toExit defaults to false if not given
+     */
+    public Command(String key, String usage, boolean toExit) {
+        this.key = key;
+        this.usage = usage;
+        this.toExit = true;
```
ZayK1: ## Potential Logic Error?
In the alternate constructor, ```toExit``` is always set to true regardless of the parameter value. Should this be ```this.toExit = toExit;```?

ZayK1 (commented):
**LGTM**, Jut a few nits to fix.

PR #154, opened 2025-08-25
[lim-james] iP

lim-james:
Dwight Clapping
# Dwight Companion

> _“Whenever I’m about to do something, I think, ‘Would an idiot do that?’ And if they would, I do not do that.” – Dwight Schrute_

The _Dwight Companion_ is the ultimate assistant for productivity and beet-based efficiency. It will not only help you manage your tasks, but also prove once and for all that computers can never outsell the best paper salesman.

* text-based like Schrute Farms invoices
* easy to learn _(even Kevin could use it)_
* FAST - To give you a reference point, it is somewhere between a snake and a mongoose… and a panther.

## How do I get Dwight as my assistant ~to the~ task manager?
* [Download it from here](https://github.com/lim-james/ip)
* Add your tasks
* Let Dwight be the best task manager you have ever seen

## Dwight's Comprehensive [Resume](https://www.youtube.com/watch?v=JJzMUqXrQrE)
 
- [X] Managing todos
- [X] Managing deadlines
- [X] Managing events
- [X] Sorting tasks

## Personal notes
It is week 6 and I have not installed IntelliJ yet.


src/main/java/duke/Duke.java
yuntaeseo, line 1: How about creating a scanner method inside Ui.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,58 @@
...
+            System.out.println("OH DEAR! " + e.getMessage());
+            return;
+        }
+
+        Scanner sc = new Scanner(System.in);
+        ui.drawLine();
+        String userInput = sc.nextLine().trim();
+        ui.drawLine();
+
+        while (!userInput.equals("bye")) {
```
yuntaeseo: How about making "bye" as an Command also, so that user input handling doesn't have to be done in Main.java

yuntaeseo (commented):
I guess simply creating a single custom exception class is possible by letting each methods that throws the custom exception to handle their own e.message

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,11 @@
...
+
+import duke.task.TaskList;
+
+public class FindCommand extends Command {
+    @Override
+    public void execute(TaskList list, String description) {
+        System.out.println("Here are the matching tasks in your list:");
+        System.out.println(list.filtered(description));
+    }
+}
```
weihan-01: Perhaps you could do the JavaDoc for this part

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,18 @@
+package duke.command;
+
+import duke.task.TaskList;
+
```
weihan-01: I like how you program is modularised into different folders like the command, storage, task etc.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Returns a serialized representation of the entire task list, suitable for saving to storage.
+     * Each task is placed on a new line.
+     *
+     * @return The serialized string of the task list.
+     */
+    public String serialize() {
+        return list.stream().map(Task::serialize).reduce("", (a, b) -> a + "\n" + b).trim();
+    }
+}
```
weihan-01: This part is cool

weihan-01 (commented):
Overall looks good to me. I like that the code is organised neatly and majority of the code has JavaDoc comments

src/main/java/duke/task/TaskList.java
lim-james: Thanks bro :)

src/main/java/duke/command/Command.java
lim-james: yessir

src/main/java/duke/command/FindCommand.java
lim-james: Good catch, I missed out upon merging the new feature. Resolved in [d7ecc8e](https://github.com/nus-cs2103-AY2526S1/ip/pull/154/commits/d7ecc8e31819a5477021a75c2fd7085d74abe797)

src/main/java/duke/Duke.java
lim-james: I like that idea actually! Will look into it

src/main/java/duke/Duke.java
lim-james: Interesting, I will consider this. But if there isn't much to be abstracted then I feel there's no point.

PR #161, opened 2025-08-25
[Khantminn290] iP

Khantminn290:
# Ronaldo ChatBot 
> Remembers your tasks and lets you edit and modify their completion 
- Fast
- Simple
- Easy to learn **commands** 

Steps to Use:
1. Download .jar file from the releases from [here](https://github.com/Khantminn290/ip/releases)
2. In your terminal `java -jar ronaldo.jar`
3. And let it manage your tasks

If you are a Java programmer, you can use it to practice Java too, Here's the `main` method:
```Java
public class Main {
        Application.launc(MainApp.class, args);
    }
}
```
Current features
- [x] Todo tasks
- [x] Deadline tasks
- [x] Event tasks
- [ ] GUI for chatbot (coming soon)

I love Ronaldo SIUUUUU! :sparkles:


src/main/java/DialogBox.java
```diff
@@ -0,0 +1,59 @@
...
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+/**
+ * Represents a dialog box consisting of an ImageView to represent the speaker's face
+ * and a label containing text from the speaker.
+ */
+public class DialogBox extends HBox {
```
jovnc, line 19: Should we put classes related to JavaFX inside a package?

src/main/java/ronaldo/parser/Parser.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param input The raw string entered by the user.
+     * @return A {@link Command} representing the action indicated by the input.
+     *         Returns {@link Command#INVALID} if the input does not match any known commands.
+     */
+    public static Command parse(String input) {
+        if (input.equals("bye")) {
+            return Command.BYE;
+        } else if (input.equals("list")) {
+            return Command.LIST;
+        } else if (input.startsWith("mark ")) {
```
jovnc, line 44: Any reason why there needs to be an extra space for `mark` and `unmark`, but it is not required by the other commands?

src/main/java/ronaldo/storage/Storage.java
```diff
@@ -0,0 +1,129 @@
...
+        ArrayList<Task> tasks = new ArrayList<>();
+        try {
+            ArrayList<String> lines = new ArrayList<>(Files.readAllLines(file));
+            for (String line : lines) {
+                String[] parts = line.split(" \\| ");
+                String type = parts[0];
+                boolean isDone = Boolean.parseBoolean(parts[1]);
+
+                switch (type) {
+                case "T":
```
jovnc: Should we extract out the magic strings "T", "D", "E" to use a named constant so that the code is easier to understand?

src/main/java/ronaldo/task/ToDos.java
```diff
@@ -0,0 +1,28 @@
+package ronaldo.task;
+
+/**
+ * Represents a To-Do task.
+ * A To-Do is a type of Task that only has a description and does not have a date/time.
+ */
+public class ToDos extends Task {
```
jovnc: Should `ToDos` be renamed to `ToDo` so that it is in singular form, similar to other task types?

src/main/java/ronaldo/ui/EmptyStringException.java
```diff
@@ -0,0 +1,14 @@
+package ronaldo.ui;
+
+/**
+ * Exception thrown when a task description is empty.
+ */
+public class EmptyStringException extends RonaldoException {
```
jovnc: Should we create a new folder to store exceptions, as it is not related to the `ui`?

src/main/java/ronaldo/ui/Ronaldo.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    /**
+     * Processes a single user input string and returns the resulting message.
+     *
+     * @param input the user input string, e.g., from userInput.getText()
+     * @return the output message to be handled elsewhere
+     */
+    public String processInput(String input) {
```
jovnc, line 100: Should we abstract some logic of the `processInput` function out, as the method is quite long and difficult for the reader to process?

src/test/java/ronaldo/task/TaskListTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.ArrayList;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class TaskListTest {
```
jovnc, line 13: I like that you have comprehensive test coverage to test most methods in the `TaskList` class, and the naming of the methods are consistent so they are easy to follow.

jovnc (commented):
Overall, I found your code to mostly follow the coding standards. Some parts of the code can be further abstracted out to avoid long methods. LGTM.

src/main/java/ronaldo/ui/Ui.java
```diff
@@ -0,0 +1,124 @@
...
+public class Ui {
+
+    /**
+     * Encloses a message in a border for visual clarity.
+     *
+     * @param message the message to encase.
+     * @return the formatted message with a border.
+     */
+    public static String encase(String message) {
+        return "____________________________________________________________\n"
```
saajid-ahamed, line 21: Could the line generation be done in a shorter way? 🤔

src/main/java/ronaldo/ui/Command.java
```diff
@@ -0,0 +1,37 @@
+package ronaldo.ui;
+
+/**
+ * Enum representing all possible commands that the user can input
+ * in the Ronaldo application.
+ */
+public enum Command {
```
saajid-ahamed: I like how you utilized Enumeration here?

src/main/java/ronaldo/storage/Storage.java
```diff
@@ -0,0 +1,129 @@
...
+
+                case "D":
+                    Deadline deadline = new Deadline(parts[2], parts[3]);
+                    if (isDone) {
+                        deadline.markAsDone();
+                    }
+                    tasks.add(deadline);
+                    break;
+
+                case "E":
```
saajid-ahamed: Do you think we can different naming of cases to make the code easier to be understood?

saajid-ahamed (commented):
LGTM. Just some nits to fix

src/main/java/ronaldo/task/ToDos.java
ryan-toh: Agree with the above, as ToDos does not represent a collection of ToDo objects.

src/main/java/ronaldo/task/Deadline.java
```diff
@@ -0,0 +1,87 @@
...
+        this.by = by;
+
+        try {
+            String[] dateTime = by.split(" ");
+            this.date = LocalDate.parse(dateTime[0], DateTimeFormatter.ofPattern("yyyy-MM-dd"));
+            this.time = LocalTime.parse(dateTime[1], DateTimeFormatter.ofPattern("HHmm"));
+        } catch (Exception e) {
+            throw new IllegalArgumentException(
+                    "Invalid deadline format. Expected format: yyyy-MM-dd HHmm", e
+            );
```
ryan-toh, line 54: I would suggest throwing a InvalidDateFormatException to be consistent.

src/main/java/ronaldo/ui/Ronaldo.java
```diff
@@ -0,0 +1,299 @@
...
+                    String by = parts[1];
+
+                    // Check date and time format yyyy-MM-dd HHmm
+                    try {
+                        java.time.format.DateTimeFormatter formatter =
+                                java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+                        java.time.LocalDateTime.parse(by, formatter);
+                    } catch (java.time.format.DateTimeParseException e) {
+                        throw new InvalidDateFormatException();
+                    }
```
ryan-toh: Perhaps you could enclose the creation of `Deadline` in a try-catch instead, as `Deadline` already verifies if the date format is incorrect (it throws an `IllegalArgumentException` as of now), so you don't have to check twice?

For example:

```
try {
 Deadline deadline = new Deadline(description, by);
} catch (InvalidDateFormatException e) {  // if Deadline throws InvalidDateFormatException instead of IllegalArgumentException
  System.out.println(e.getMessage);
}
```

src/main/java/ronaldo/task/Event.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Constructs an Event task with a description, start time, and end time.
+     *
+     * @param description the description of the event
+     * @param from the start time of the event
+     * @param to the end time of the event
+     */
+    public Event(String description, String from, String to) {
+        super(description);
```
ryan-toh, line 23: I would recommend adding some validation logic to ensure that the date is entered in the correct format (yyyy-MM-dd HHmm) similar to Deadline.

src/main/java/ronaldo/ui/Ronaldo.java
```diff
@@ -0,0 +1,299 @@
...
+                }
+                String by = parts[1];
+
+                try {
+                    java.time.format.DateTimeFormatter formatter =
+                            java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+                    java.time.LocalDateTime.parse(by, formatter);
+                } catch (java.time.format.DateTimeParseException e) {
+                    throw new InvalidDateFormatException();
+                }
```
ryan-toh: You can apply the changes I suggested with regards to enclosing `Deadline` in a try-catch so that you can remove this as well!

src/main/java/ronaldo/ui/Ui.java
ryan-toh, line 21: Great use of `encase()` so you don't have to type this out every time you create a new CLI output function.

ryan-toh (commented):
Looks well organised! Just a few suggestions with regards to `Event`, and some changes to exception handling throughout. LGTM.

PR #165, opened 2025-08-25
[pohanson] iP

pohanson:
# Bob, building a better you

> "With **Bob**, I have _never forgotten an event, task, or todo._" - A satisfied user 😄 

How to get it:

1. Download the jar [here](https://github.com/pohanson/ip/releases/download/A-Jar/Bob.jar)
2. `java -jar <filename>` to run the file
3. Just use it.

## Commands

It allows you to perform this actions via **text commands**:

* `list` - List all tasks
* `mark <task number>` - Mark a task as done
* `unmark <task number>` - Unmark a task as not done
* `todo <description>` - Add a task without any date/time attached to it
* `deadline <description> /by <datetime>` - Add a task that need to be done before a specific date/time
* `event <description> /from <start datetime> /to <end datetime>` - Add task that start at a specific date/time and ends at a specific date/time
* `delete <task number>` - Delete a task from the list
* `find <search keyword>` - Search for a task that contains <search keyword> in the description.
* `bye` - Exit the program

## Features:

- [x]  Mange Todo
- [x]  Mange Deadline
- [x]  Mange Event
- [ ]  Nice UI

Practice Java if you're a Programmer, below is the entry point of the code. It takes a string which is the _file path_
```java
    /**
     * Entry point for the app.
     */
        new Bob("data/tasklist.txt").run();
    }
```


src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,67 @@
...
+            tasks = new TaskList(this.storage);
+        } catch (InvalidInputException e) {
+            ui.showError("Error loading tasks from file, using empty task list: " + e.getMessage());
+            tasks = new TaskList(this.storage);
+        }
+
+    }
+
+    /**
+     * Main run loop for the application.
```
Joulenergy:
```suggestion
     * Runs the main loop for the application.
```
Based on the coding standard, the first sentence of method header comments should start in the form `Returns ...`, `Sends ...`, `Adds ...`

src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,67 @@
...
+            command.execute(tasks, ui, storage);
+            if (command instanceof ExitCommand) {
+                break;
+            }
+        }
+        scanner.close();
+    }
+
+    /**
+     * Entry point for the app.
```
Joulenergy:
```suggestion
     * Starts the Bob application
```
Same comment as above

src/main/java/bob/task/TaskList.java
```diff
@@ -0,0 +1,163 @@
...
+        return t;
+    }
+
+    /**
+     * Validates the task index.
+     *
+     * @param index the 0-based index of the task
+     * @return true if the index is valid, else false.
+     */
+    public boolean validateTaskIndex(int index) {
```
Joulenergy: Perhaps can rename to sound more like boolean? For example isValidTaskIndex()

src/main/java/bob/task/ToDos.java
```diff
@@ -0,0 +1,41 @@
+package bob.task;
+
+import bob.exception.InvalidInputException;
+
+/**
+ * Tasks without any date/time attached to it.
+ */
+public class ToDos extends Task {
```
Joulenergy:
```suggestion
public class Todo extends Task {
```
Perhaps it would be good to rename it to either Todo or ToDo? (Non plural form)

src/main/java/bob/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+
+import bob.exception.InvalidInputException;
+import bob.task.Task;
+import bob.task.TaskList;
+
+/**
+ * Storage handles loading tasklist from a file and saving it.
+ */
+public class Storage {
+    private final String filePath;
```
Joulenergy, line 21: Is there a reason filePath was not declared as a constant? (private static final FILE_PATH)

src/main/java/bob/task/Deadlines.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+import bob.exception.InvalidInputException;
+
+/**
+ * Tasks that need to be done before a specific date/time.
+ */
+public class Deadlines extends Task {
```
Joulenergy:
```suggestion
public class Deadline extends Task {
```
Consider renaming the class from `Deadlines` to `Deadline` since each instance represents a single task with a specific deadline?

src/main/java/bob/task/Events.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+import bob.exception.InvalidInputException;
+
+/**
+ * Tasks that start at a specific date/time and ends at a specific date/time.
+ */
+public class Events extends Task {
```
Joulenergy:
```suggestion
public class Event extends Task {
```
Since the class represents a single entity, I find Event more intuitive, what do you think?

src/main/java/bob/command/FindCommand.java
```diff
@@ -0,0 +1,26 @@
...
+
+import bob.storage.Storage;
+import bob.task.TaskList;
+import bob.ui.Ui;
+
+/**
+ * Command to find a task by description
+ */
+public class FindCommand extends Command {
+    private final String input;
```
Joulenergy, line 14:
```suggestion
    /** User input string to search for. */
    private final String input;
```
Javadoc of class members could be added as well

Joulenergy (commented):
Looks good 👍 Overall, I found your code easy to read. I noted a few minor coding standard violations and nits

PR #174, opened 2025-08-25
[Marcus Ng Zheng Han] iP

PeanutButter1212:
### Peanut

Peanut helps you keep track of deadlines and tasks to complete, it is:

- text-based
- easy to learn
- SUPER FAST to use

All you need to do is,

1. download it from [[here](https://github.com/PeanutButter1212/ip)]
2. double-click it.
3. add your tasks.

Features:
- [x] Managing tasks
- [x]  Managing deadlines 
- [ ] Reminders (coming soon)

```
public class Main {
        Application.launch(Peanut.class, args);
    }
}
```


src/main/java/peanut/PeanutException.java
```diff
@@ -0,0 +1,8 @@
+package peanut;
+
+public class PeanutException extends Exception {
+    public PeanutException(String a) {
```
zhiyu05: Perhaps a more intuitive parameter name than `a` (eg. `message`) would make this clearer? That way it is more consistent with `Exception`

src/main/java/peanut/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            String[] sep = parts[1].split("/by", 2);
+
+            if (sep.length < 2 || sep[0].isBlank() || sep[1].isBlank()) {
+                throw new PeanutException("The description/time of deadline cannot be empty!! "
+                        + "(e.g deadline read book /by 2019-10-15)");
+            }
+
+            String task = sep[0];
```
zhiyu05: Could this variable name `task` be confusing? It seems to hold the description rather than the full task. Maybe `taskDescription` would be clearer.

src/main/java/peanut/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+
+        if (userInput.startsWith("event")) {
+            String[] parts = userInput.split("\\s+", 2);
+
+            if (parts.length < 2 || parts[1].isBlank()) {
+                throw new PeanutException("The description/time of start and deadline "
+                        + "cannot be empty!! (e.g event project meeting /from 2019-10-15 /to 2019-10-16)");
+            }
+
+            String[] sep = parts[1].split("/from", 2);
```
zhiyu05: Any reason you went with names like `sep` and `sep2`? I wonder if more descriptive names (eg. descriptionAndStart, startAndDeadline) would make the parsing logic easier to follow.

src/main/java/peanut/Task.java
```diff
@@ -0,0 +1,66 @@
+package peanut;
+
+/**
+ * Represents a task that user inputs
+ * Provides methods to display, mark, and unmark tasks.
+ */
+public class Task {
```
zhiyu05: I like how you have separated the logic of tasks to a separate class for OOP.

src/main/java/peanut/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+            return true;
+        }
+
+        if (userInput.equals("list")) {
+            ui.showListMessage(taskList);
+            return false;
+        }
+
+        if (userInput.startsWith("mark")) {
+            String[] parts = userInput.split(" ", 2);
```
zhiyu05: I have noticed this validation pattern appearing several times throughout this method. Perhaps you could create a helper function (eg.validateCommandHasArguments()) to reduce code duplication?

src/main/java/peanut/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+     * @param userInput Command that user enters.
+     * @param taskList TaskList that is loaded from previously saved file
+     * @param ui UI to handle user instructions and display messages
+     * @param storage Storage to handle loading and saving of files
+     * @throws PeanutException If the input cannot be translated into a valid command
+     */
+    public boolean parse(String userInput, TaskList taskList, Ui ui, Storage storage) throws PeanutException {
+        userInput = userInput.trim();
+
+        if (userInput.equals("bye")) {
```
zhiyu05: I have noticed you use a long chain of if-statements to handle multiple different inputs, which is handled well. Perhaps you could also consider using a switch statement which might make adding new commands easier.

src/main/java/peanut/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+        return tasks.size();
+    }
+    /**
+     * Delete task at the specified index.
+     *
+     * @param taskNumber Index of task that user wants to delete.
+     */
+    public void delete(int taskNumber) {
+        tasks.remove(taskNumber);
+    }
```
TeoShiYuan: The param used here is taskNumber, but other methods (mark, unmark) use index. For consistency, would it be better to standardize the naming?

src/main/java/peanut/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param index Index of task that user wants to unmark.
+     */
+    public void unmark(int index) {
+        tasks.get(index).unmark();
+    }
+    /**
+     * Returns number of tasks in the TaskList
+     *
+     * @return The number of tasks in the Tasklist
+     */
```
TeoShiYuan: Minor typo here: Tasklist -> TaskList

src/main/java/peanut/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+        List<Task> matches = new ArrayList<>();
+        for (Task t : tasks) {
+            if (t.getDescription().contains(keyword)) {
+                matches.add(t);
+            }
+        }
+        return matches;
+    }
+
+}
```
TeoShiYuan: Well done, easy to read and understand the logic of the method.

Suggestion: contains is case-sensitive. You could consider using a case-insensitive comparison, to mark it more user friendly:

`if (t.getDescription().toLowerCase().contains(keyword.toLowerCase())) {`

src/main/java/peanut/Storage.java
```diff
@@ -0,0 +1,116 @@
...
+     * Loads task from file into a list of tasks
+     * If file/folder does not exist, empty list is returned
+     *
+     * @return A list of task loaded from the file/empty list
+     */
+    public List<Task> load() {
+        File folder = new File("./data");
+        if (!folder.exists()) {
+            folder.mkdirs();
+        }
```
TeoShiYuan: Nice to ensure that the ./data folder exists before attempting to load tasks.

Question: Would it make sense to extract this repeated folder creation logic into a private helper method (since the same code appears in both `load()` and `save()`)?

src/main/java/peanut/Storage.java
```diff
@@ -0,0 +1,116 @@
...
+            break;
+        case "D":
+            task = new Deadline(desc, parts[3]);
+            break;
+        case "E":
+            task = new Event(desc, parts[4], parts[3]);
+            break;
+        default:
+            return null;
+        }
```
TeoShiYuan: The order of parts[4] coming before parts[3] seems unintuitive.
I noticed that in your Event constructor, deadline comes before the start date. Perhaps you might want to switch the parameters for the Event constructor?

PR #207, opened 2025-08-25
[MinhMXC] iP

ngqaminh:
### MinhGPT

> Live more, dementia less

MinhGPT cures your dementia, allowing you to remember what you need to do. It is:
- command line based
- easy to use
- *EXTREMELY* EFFICIENT

All you need to do is:
1. Download it from [here](https://github.com/MinhMXC/ip/releases).
2. Double-click it.
3. Add your tasks.
4. Dementia cured. 😊 

And it is **OPEN-SOURCE** & **FREE**!

Features:
- [x] Manage tasks
- [ ] Manage deadlines (in testing)
- [ ] Reminders (in testing)

```java
public class Main {
    }
}
```


src/main/java/minhgpt/util/Logger.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Print a log message at level 'Error'
+     */
+    public static void error(String message) {
+        // Red background
+        String errorBg = "\033[41m";
+        System.out.println(String.format("%s%s ERROR %s %s", TEXT_COLOR, errorBg, RESET, message));
+    }
+}
```
limweekit: Perhaps you could rename these methods to printInfo, printWarning, printError to make their names clearer as to what they do?

src/main/java/minhgpt/ui/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Return the response when user add a task.
+     *
+     * @param task Task that was just added by user.
+     */
+    public String addResponse(Task task) {
+        return String.format("(˶ᵔ ᵕ ᵔ˶) Added: %s", task);
+    }
+
```
limweekit: I love the flair UI responses, really showcases the personality of the chatbot.

src/main/java/minhgpt/task/TaskList.java
```diff
@@ -0,0 +1,72 @@
...
+    public int size() {
+        return tasks.size();
+    }
+
+    /**
+     * Return the task at index 'index' in list.
+     */
+    public Task get(int index) {
+        return tasks.get(index);
+    }
```
limweekit: Perhaps getIndex would be a more descriptive name?

limweekit (commented):
Good job on the iP so far. The code is well structured and follows closely to the naming conventions of the course, with a few nits here and there. I really like the personality of the chatbot and the use of gifs is very creative as well. Keep up the good work.

src/main/java/minhgpt/util/Logger.java
ngqaminh: Thank you for pointing this out. I forgot I even had this logging class haha.

PR #209, opened 2025-08-26
[shaneak03] iP

shaneak03:
# Barry ⚡ 
> "Things aren't always what they seem. Our fears can play tricks on us... But usually, hidden behind our fears are second chances waiting to be seized." – *Barry Allen*

Barry can help you with:
- Adding new tasks  
- Deleting completed tasks  
- Listing all existing tasks  
- Having fun chatting with a **speedster**  

Here’s how you can get started with Barry:
1. Clone the [repo](https://github.com/shaneak03/ip)
2. Ensure you have the **JDK** and necessary packages installed  
3. Run ***Barry*** and ~~procrastinate~~ *zoom* into productivity 💨  

```java
public class Barry {
    private String name;

    public Barry(String s) {
        this.name = s;
    }

    @Override
    public String toString() {
        return "Hey, I'm " + this.name + "! Ready to help at lightning speed.";
    }
}
```
### Try the following `Task List`
- [x] Buy bread  
- [ ] Investigate a crime scene
- [ ] Call Barry for backup


src/main/java/barry/Storage.java
```diff
@@ -0,0 +1,153 @@
...
+        if (parentDir != null && !parentDir.exists()) {
+            parentDir.mkdirs();
+        }
+    }
+
+    /**
+     * Appends tasks to file
+     * @param task the task to save
+     */
+    public void save(Task task) {
```
noahseethorcodes: Would it make sense to rename the method from ```save``` to ```saveTask``` so that it’s maximally clear what the function does? Right now, ```save``` could be interpreted as saving multiple tasks or something more general, whereas ```saveTask``` makes the intent explicit.

src/main/java/barry/Storage.java
```diff
@@ -0,0 +1,153 @@
...
+            writer.write(task.toSaveString() + System.lineSeparator());
+        } catch (IOException e) {
+            System.out.println("Error saving task: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Clears all tasks in the storage file
+     */
+    public void delete() {
```
noahseethorcodes: Would it make sense to rename the method from ```delete``` to ```clear``` or ```deleteAllTasks``` so that it’s maximally clear what the function does? Right now, ```delete``` could be interpreted as deleting certain tasks or something more general, whereas ```deleteAllTasks``` makes the intent explicit.

src/main/java/barry/Storage.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+        } catch (IOException e) {
+            System.out.println("Error initializing storage file: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Creates parent directories if they don't exist
+     */
+    private void ensureParentDirExists() {
```
noahseethorcodes: Honestly a nit, but might you want to spell out Directory just for clarity?

src/main/java/barry/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+        System.out.println(message);
+        showLine();
+    }
+
+    /**
+     * Reads the next command entered by User
+     * @return raw input string by user
+     */
+    public String readCommand() {
+        System.out.print("User: \n");
```
noahseethorcodes: This ```User: ``` prefix is nice! You've inspired me to add something similar to my own project

noahseethorcodes (commented):
Code looks good! Just some nits regarding certain methods names.

src/main/java/barry/Barry.java
```diff
@@ -0,0 +1,200 @@
...
+                    ui.showLine();
+                    continue;
+                }
+            }
+
+            ui.showLine();
+
+            try {
+                switch (command.toLowerCase()) {
+                case "bye":
```
jongrazi: I think case here should be indented

jongrazi (commented):
LGTM, Overall code looks neat and easily readable and understood.

PR #217, opened 2025-08-26
[noahseethorcodes] ip

noahseethorcodes:
# Logos iP Pull Request

Welcome to my **Logos** iP submission! This PR contains the core chatbot features for task management.
I’ve included the following elements:

---

## Features Implemented

* Parse user input into commands (`todo`, `deadline`, `event`, etc.)
* Store tasks in memory and in a local text file
* Display tasks with a clean UI box format
* Handle errors gracefully with custom exceptions

---

## Key Milestones

1. Refactor into multiple packages (`ui`, `parser`, `storage`, `tasklist`)
2. Implement `Command` pattern for extensibility
3. Add support for date/time parsing with `LocalDateTime`
4. Stretch goals: `due <date>` query, reminders

---

## Example Usage

```java
// Example: adding a deadline
deadline return book /by 2019-12-02 1800
```

---

## Task List

* [x] Set up Gradle project
* [x] Add basic task types (`Todo`, `Deadline`, `Event`)
* [x] Implement `mark` and `unmark`
* [ ] Implement JavaFX for GUI
* [ ] Write more tests with JUnit 5

---

## Notes

> "Good code is like a good joke — it needs no explanation."
> — *Martin Fowler (probably)*

---

## Links

* Project README: [[Skeleton Task Manager](https://github.com/nus-cs2103-AY2425S1/ip)]
* Inline reference: run the program with `./gradlew run`

---

## Formatting Examples

* This project is in **Java**
* User input is parsed using *regular expressions*
* Some old code has been ~~removed~~ refactored for clarity

---

That’s all! Looking forward to feedback.


src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
+package ui;
+
+import java.util.List;
+import java.util.Scanner;
+
+public class Ui {
+    private final int INDENT_LENGTH = 4;
+    private final int LINE_LENGTH = 80;
```
limanjun99: Should these two fields be static? The style guide says to use screaming snake case for constants, and links to [Google Java style guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names), which seems to define constants as static final fields.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+    private final Scanner sc = new Scanner(System.in);;
+
+    public String readLine() {
+        String userInput = sc.nextLine();
+        return userInput;
+    };
```
limanjun99: Should not have a semicolon after function definition according to style guide. I noticed the same issue in several other places too.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private void closeScanner() {
+        sc.close();
+    };
+
+    // Helper Function that indents text and wraps with horizontal lines
```
limanjun99: Should this be a JavaDoc comment instead? E.g. something with the following form:

```suggestion
/**
 * Helper function that indents text and wraps with horizontal lines.
 * /
```

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
+package ui;
+
+import java.util.List;
+import java.util.Scanner;
+
+public class Ui {
+    private final int INDENT_LENGTH = 4;
+    private final int LINE_LENGTH = 80;
+    private final Scanner sc = new Scanner(System.in);;
```
limanjun99: Nit: Extra semicolon.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public void showWelcome(String chatbotLogo, String chatbotName) {
+        System.out.println("Welcome to...\n" + chatbotLogo);
+        this.respond(
+                "Hello! I'm " + chatbotName + " :) Your friendly terminal task manager.",
+                "",
+                "Here are some commands you can try:",
+                "-> todo <desc>                             : Add a simple task",
+                "-> deadline <desc> /by <time>              : Add a task with a deadline (<yyyy-MM-dd HHmm>)",
+                "-> event <desc> /from <start> /to <end>    : Add an event with a start and end time (<yyyy-MM-dd HHmm>)",
```
limanjun99, line 89: This line is 122 characters long, would it be possible to tweak it / format it slightly differently to fit within the hard limit of 120 characters per line?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.List;
+import java.util.Scanner;
+
+public class Ui {
+    private final int INDENT_LENGTH = 4;
+    private final int LINE_LENGTH = 80;
+    private final Scanner sc = new Scanner(System.in);;
+
+    public String readLine() {
```
limanjun99, line 29: Since this is a public method, would it make sense to add some JavaDoc comments explaining what this method does? I noticed several other public methods (in other classes too) also lack JavaDoc comments.

src/main/java/logos/Logos.java
```diff
@@ -0,0 +1,67 @@
...
+
+import errors.UnknownCommandException;
+import localstorage.Storage;
+import parser.Parser;
+import tasklist.TaskList;
+import errors.InvalidCommandFormatException;
+import errors.InvalidIndexException;
+import errors.LogosException;
+import commands.ByeCommand;
+import commands.Command;
```
limanjun99, line 15: The style guide mentions that import ordering should be consistent (I think a common one is lexicographical ordering). Just curious, is that guideline being followed here? As this doesn't seem to be lexicographical (though any other ordering is probably fine too).

src/main/java/logos/Logos.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class Logos {
+    private static TaskList taskList;
+
+    private static String LOCAL_STORAGE_FILE_PATH = "./data/tasks.txt";
```
limanjun99, line 19: Same question as the one in `Ui.java`, should this constant be final too in addition to being static?

src/main/java/logos/Logos.java
```diff
@@ -0,0 +1,67 @@
...
+            } catch (InvalidCommandFormatException e) {
+                ui.respond(e.getMessage());
+            } catch (InvalidIndexException e) {
+                ui.respond(e.getMessage());
+            } catch (IOException e) {
+                ui.respond("Error handling local storage: " + e.getMessage());
+            } catch (LogosException e) {
+                ui.respond(e.getMessage());
+            }
+        }
```
limanjun99, line 65: I think the line breaks + useful comments to separate logical sections here are very nice :), makes for readable code.

src/main/java/tasklist/TaskList.java
```diff
@@ -0,0 +1,106 @@
...
+        storage.saveTasks(tasks);
+    }
+
+    public List<String> listTasks() {
+        return tasks.stream()
+                .map(Task::getAsListItem)
+                .toList();
+    }
+
+    public Todo addTodo(String taskName) throws IOException{
```
limanjun99: Nit: Whitespace before `{`

```suggestion
    public Todo addTodo(String taskName) throws IOException {
```

limanjun99 (commented):
LGTM, just a few minor code style issues.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Will update to keep consistent with the styleguide.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Will audit to keep consistent with the styleguide.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Will update to be a JavaDoc.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Will remove.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Will update accordingly.

src/main/java/ui/Ui.java
noahseethorcodes: Thanks for the spot! Still in the midst of writing JavaDocs for the public methods.

src/main/java/logos/Logos.java
noahseethorcodes: Nope. Currently not much thought has been put into the package import ordering. Even from this code snippet you highlighted there are imports for the same package but separated. Will re-organise them!

src/main/java/logos/Logos.java
noahseethorcodes: Thanks for the spot! Will update to be final.

src/main/java/tasklist/TaskList.java
noahseethorcodes: Thanks for the spot! Will update accordingly.

PR #229, opened 2025-08-26
[Hao Rui] iP

haorui2004:
# Nina
> Work hard today, nap harder tomorrow

Nina is more than just a chatbot, she is your **new personal assistant** 😉
- Text-based
- Easy to learn
- ~~FAST~~SUPER FAST to use

## You can start by
1. Download the project
2. Double click it
3. Add your tasks
4. Let Nina manage your tasks for you!

## Features
- [x] Add/Delete tasks
- [x] Mark/Unmark done
- [x] Handle dates in tasks
- [x] Print the whole list
- [ ] User-friendly UI (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the ```main``` method:
```
        Application.launch(Main.class, args);
    }
```
This project is built based on [iP](https://github.com/nus-cs2103-AY2526S1/ip/tree/master#) from CS2103T AY2526S1


src/main/java/DialogBox.java
```diff
@@ -0,0 +1,57 @@
...
+import javafx.collections.ObservableList;
+import javafx.fxml.FXML;
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Pos;
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
```
udonzzz: Hi it seems that these import statements are duplicated from the ones in lines 4 and 5, of which the ones in lines 4 and 5 are in the correct order for import statements

src/main/java/nina/CommandException.java
```diff
@@ -0,0 +1,7 @@
+package nina;
+
+public class CommandException extends Exception{
```
udonzzz: Hi seems like a whitespace between Exception and { has been missed out? Similarly for InvalidInputException.java

src/main/java/nina/Nina.java
```diff
@@ -0,0 +1,88 @@
+package nina;
+
+import nina.command.Command;
+import nina.task.TaskList;
+
+import java.util.Scanner;
```
udonzzz: Hi I think java.util.Scanner should be imported before your custom imports as per the import statement order

src/main/java/nina/Parser.java
```diff
@@ -0,0 +1,96 @@
+package nina;
+
+import nina.command.*;
```
udonzzz: While its true that you have many command classes, I believe that we are not supposed to use wildcard imports, and classes should be imported explicitly

src/main/java/nina/command/DeleteCommand.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class DeleteCommand implements Command {
+    protected int index;
+
+    /**
+     * Constructs a DeleteCommand object.
+     *
+     * @param index index of the task to be deleted
+     */
+
```
udonzzz: Hi I think you might have accidentally added an empty line between the comment and the DeleteCommand method?

src/main/java/nina/command/MarkCommand.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class MarkCommand implements Command {
+    protected int index;
+
+    /**
+     * Constructs a MarkCommand object.
+     *
+     * @param index index of the task to be marked
+     */
+
```
udonzzz: Hi it seems you might have accidentally added an empty line here too between the comment and your method, since you did not do so for your other methods and their comments

udonzzz (commented):
Hi I think your code is really nice! No big glaring issues honestly, just maybe some small coding standard violations here and there.

src/main/java/nina/Nina.java
```diff
@@ -0,0 +1,88 @@
...
+            } catch (CommandException e) {
+                ui.showError("nina.command.Command error: " + e.getMessage());
+            } catch (InvalidInputException i) {
+                ui.showError("Input error: " + i.getMessage());
+            }
+        }
+    }
+
+    public String getResponse(String input) {
+        String trimmed = input.trim();
```
Zisaac99: Perhaps this could be renamed to trimmedInput for clarity?

src/main/java/nina/task/Task.java
```diff
@@ -0,0 +1,120 @@
...
+    protected String basePrefix(String type) {
+        return type + " | " + (isDone ? "1" : "0") + " | " + description;
+    }
+
+    /**
+     * Converts this task into a save line string for storage.
+     *
+     * @return the save line string representation of the task
+     */
+    public abstract String toSaveLine();
```
Zisaac99: Perhaps this could be named getSavedLine?

src/main/java/nina/task/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+
+/**
+ * Represents a list of Task objects.
+ */
+public class TaskList implements Serializable {
+    protected ArrayList<Task> tasks;
```
Zisaac99, line 11: Nice naming of your arraylist.

Zisaac99 (commented):
Overall looks good

src/main/java/nina/Parser.java
haorui2004: Oh I actually wrote them all but I think my IDE just folded them😱. Thanks and I will check again!

src/main/java/nina/Nina.java
haorui2004: Okk thank you

src/main/java/nina/task/Task.java
haorui2004: oh actually this one is to convert a task into a save-able string. Maybe I will change to toSavableLIne? Thanks

PR #237, opened 2025-08-26
[kevintanjc] iP

kevintanjc:


src/main/java/resources/util/services/LoadingService.java
```diff
@@ -0,0 +1,200 @@
...
+     * Creates an EventTask from a given line.
+     * @param line  the {@code String} from the file representing an EventTask.
+     * @return {@link EventTask} - the created EventTask object.
+     */
+    private EventTask createEventTask(String line) {
+        String[] parts = line.split(" \\(from: | to: |\\)");
+        String description = parts[0].substring(7);
+        String startDate = parts[1];
+        String endDate = formatDate(parts[2]);
+        boolean completed = checkCompletedTask(parts[0].substring(3));
```
lihloway: would isCompleted be better? completed is also used many more times throughout

src/main/java/resources/util/services/BotService.java
```diff
@@ -0,0 +1,258 @@
...
+     * based on certain conditions.
+     *
+     * @param taskFlag  an Integer representing the type of task (1 for To-Do, 2 for Deadline, 3 for Event).
+     * @param inputString the input String containing the task details.
+     * @param checkList the Checklist object to which the task will be added.
+     * @throws IllegalStateException    if an invalid task type is provided.
+     * @throws NullPointerException     if task creation fails due to null values.
+     * @return
+     */
+    private void insertTaskIntoChecklist(Integer taskFlag, String inputString, CheckList checkList)
```
lihloway: Would something like taskType make more sense since I think flags are generally booleans?

inputString could also just be input, since the type is already specified as String

src/main/java/resources/util/datastorage/CheckList.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Represents a checklist that manages a list of tasks.
+ * <p>
+ * The {@link CheckList} class provides methods to handle and manipulate a list of {@link Task} objects.
+ *
+ * @author Kevin Tan
+ */
+public class CheckList {
+
+    private List<Task> checkList;
```
lihloway, line 18: I believe checklist (all lowercase) instead of checkList is used many times later on, which makes more sense.

Maybe use the same one to be consistent?

lihloway (commented):
This is my 2nd PR, so I looked mostly at the naming of variables.
It all looks really nice to read and makes sense, only some nitpicks from me. Nice! 🔥

src/main/java/resources/util/services/LoadingService.java
kevintanjc: Thanks for spotting this 👍

src/main/java/resources/util/services/BotService.java
kevintanjc: Thank you for your insights! 🐱

src/main/java/resources/util/datastorage/CheckList.java
kevintanjc: Will do! 😆

src/main/java/resources/util/constants/BotConstants.java
BoredFurry, line 1: I like the idea of having a separate class that stores the constant that the bot needs. Just a question but what were considerations for this as opposed to letting the relevant classes provide the information as needed?

src/main/java/resources/util/services/BotService.java
BoredFurry, line 1: What was the thought behind using a large function to handle the execution of commands as opposed to more specific interfaces and subclasses? This leads to this botservice class performing a lot of functions where I have chosen to offload to a helper class.

src/main/ui-resources/images/bart.png
BoredFurry, line 1: ybm97twmjtsz

PR #242, opened 2025-08-26
[Yap Ze Kai] iP

zekaistic:
# Chungus
> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes/))

Chungus frees your mind of having to remember things you need to do. It's,
- test-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use

All you need to do is,
1. download it from [here](https://nus-cs2103-ay2526s1.github.io/website/schedule/week4/project.html)
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---

```
public class Main {
    }
}
```


src/main/java/chungus/Chungus.java
```diff
@@ -0,0 +1,255 @@
+package chungus;
+
+/**
+ * Main application entry point for the Chungus task manager.
+ */
```
enxi-low: Perhaps the javadoc header could be right above the class?

src/main/java/chungus/Chungus.java
```diff
@@ -0,0 +1,255 @@
...
+        java.util.ArrayList<Task> matches = tasks.findByKeyword(keyword);
+        ui.showFindResults(matches);
+    }
+
+    /**
+     * Deletes a task at the index specified in the input.
+     *
+     * @param input Raw user input containing the task index.
+     * @throws ChungusException if the task index is invalid.
+     */
```
enxi-low: I like how you have javadoc comments for private methods as well.

src/main/java/chungus/CommandType.java
```diff
@@ -0,0 +1,47 @@
...
+
+    CommandType(String command) {
+        this.command = command;
+    }
+
+    public String getCommand() {
+        return command;
+    }
+
+    public static CommandType fromInput(String input) {
```
enxi-low: Perhaps this method could start with a verb?

enxi-low (commented):
LGTM. Coding standard is mostly followed.

src/main/java/chungus/Chungus.java
```diff
@@ -244,10 +244,219 @@ private void addEvent(String input) throws ChungusException {
...
+                    return ""; // ignore silently
+                } else {
+                    throw new ChungusException("I'm sorry, but I don't know what that means :-(");
+                }
+            } else {
+                return processCommandForGui(command, input);
+            }
+        } catch (ChungusException e) {
+            return ui.getErrorMessage(e.getMessage());
+        } catch (Exception e) {
```
SAKICHANNN: Hi! I would not recommend broad catch of Exception as it hides the real error during debugging and testing, would recommend removing the broad catch of Exception and let the program crash if there really is one based on instructions of CS2103. If we insist on catching it, I would suggest throwing a RuntimeException and throw e instead.

SAKICHANNN:
```suggestion
```

src/main/java/chungus/Chungus.java
```diff
@@ -244,10 +244,219 @@ private void addEvent(String input) throws ChungusException {
...
+    public String getResponse(String input) {
+        if (input.trim().equals(CommandType.BYE.getCommand())) {
+            saveTasksToStorage();
+            return ui.getByeMessage();
+        }
+        try {
+            CommandType command = parser.parseCommandType(input);
+            if (command == null) {
+                if (input.trim().isEmpty()) {
+                    return ""; // ignore silently
```
SAKICHANNN: I wonder why we are ignoring the empty input silently

src/main/java/chungus/Chungus.java
```diff
@@ -244,10 +244,219 @@ private void addEvent(String input) throws ChungusException {
...
+     * @param input      Raw user input containing the task index.
+     * @param markAsDone True to mark as done; false to mark as not done.
+     * @return Formatted response string
+     * @throws ChungusException if the task index is invalid.
+     */
+    private String markTaskForGui(String input, boolean markAsDone) throws ChungusException {
+        int idx = parser.parseTaskIndex(input,
+                markAsDone ? CommandType.MARK.getCommand() : CommandType.UNMARK.getCommand());
+        if (idx < 0 || idx >= tasks.size()) {
+            throw new ChungusException("Invalid task number. Please enter a number between 1 and " + tasks.size());
```
SAKICHANNN: In case tasks.size() returns 0, the error message would be rather confusing by saying Please enter a number between 1 and 0.

SAKICHANNN (commented):
Hi, I would recommend pushing code in a more fragmented way as suggested by the CS2103 instructions. Doing so would make it easier to review, also easier to revert when some issue arise.

PR #243, opened 2025-08-26
[miinren] iP

miinren:
### Juin
Juin helps you keep track of deadlines and tasks to complete, it is:

1. text-based
2. easy to learn
3. SUPER FAST to use

All you need to do is,

1. download it from [[here](https://github.com/miinren/ip)]
2. double-click it.
3. add your tasks.

Features:

- [x]  Managing tasks
- [x]  Managing deadlines
- [ ]  Reminders (coming soon)

```
public class Main {
        Application.launch(Juin.class, args);
    }
}
```


aliensarefake (commented):
## Code Review - Coding Standards and Quality

### Issues Found:

1. **Package Naming Inconsistency** 
   - In `build.gradle`: `mainClass.set("PeanutButter.Juin")` uses PascalCase
   - Actual package uses lowercase: `package peanutbutter;`
   - This will cause runtime errors. Should be `peanutbutter.Juin` in build.gradle

2. **JavaFX Version Management**
   - Declared `javaFxVersion = '17.0.7'` but hardcoded `24.0.1` in dependencies
   - Should use: `implementation "org.openjfx:javafx-base:${javaFxVersion}:mac"`

3. **Platform-Specific Dependencies**
   - All JavaFX deps use `:mac` classifier, limiting cross-platform compatibility
   - Consider removing platform classifier or making it configurable

### Positive Observations:
- Excellent JavaDoc documentation throughout
- Clean Command pattern implementation  
- Good separation of concerns with package structure
- Proper use of abstract classes and inheritance

yjjr0 (commented):
Not bad. For the UI class, i recommend to include an arraylist to store the lastMessage of each command instead of a string. Even better to implement another class to add and get the next / prev commands.

src/main/java/peanutbutter/Juin.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+    }
+
+    /**
+     * Generates a response for the user's chat message.
+     */
+    public String getResponse(String input) {
+        try {
+            Command cmd = Parser.parse(input);
```
Glacialthorn, line 75: I like the use of a class to deal with the different commands that a user could input!

src/main/java/peanutbutter/MainWindow.java
```diff
@@ -0,0 +1,53 @@
...
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    private Button sendButton;
+
+    private Juin juin;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/DaUser.png"));
+    private Image dukeImage = new Image(this.getClass().getResourceAsStream("/images/DaDuke.png"));
```
Glacialthorn: Maybe dukeImage could be renamed to juinImage? so that it fits with the name of the bot

Glacialthorn (commented):
LGTM except for a few nits!

PR #245, opened 2025-08-26
[KH-boop-bit]  iP

KH-boop-bit:
**Jamal**
> “Your mind is for having ideas, not holding them.” – David Allen

Jamal frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here] (https://github.com/KH-boop-bit/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😄 

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```Java
public class Jamal {
        new Jamal("data/jamal.ui.Jamal.txt").run();
    }
}
```


src/main/java/jamal/command/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
+package jamal.command;
+
+import jamal.task.TaskList;
+import jamal.util.Storage;
+import jamal.ui.Ui;
```
forestmu: Not a big issue but maybe you want to put them in order? Some are in order while some are not so it might be better to standardise.
```
import jamal.util.Storage;
import jamal.ui.Ui;
``` 
to 
```
import jamal.ui.Ui;
import jamal.util.Storage;
```

src/main/java/jamal/command/DeadlineTaskCommand.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Deadline Command SubClass for deadline task creation
+ */
+
+public class DeadlineTaskCommand extends Command {
+
+    protected Deadline task;
+
+    public DeadlineTaskCommand(Task task) {
```
forestmu, line 16: you should add a comment for this (and the following classes) since it is public

src/main/java/jamal/command/DeadlineTaskCommand.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * Executes Command on TaskList
+     * Writes an unmarked Deadline data into Storage
+     *
+     * @param taskList Tasklist that contains an Arraylist of tasks
+     * @param ui User interface for print statements
+     * @param storage Stores data and allow read write operations on it
+     */
```
forestmu, line 27: Nice to not omit this comment for this and following command classes👍

src/main/java/jamal/task/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+    protected LocalDateTime by;
+
+    /**
+     * Deadline task creation
+     *
+     * @param description Task description
+     * @param by Deadline date
+     * @throws DateTimeParseException if by is not of parsable date format "yyyy-mm-ddThh:mm:ss"
+     */
+    public Deadline(String description, String by) throws DateTimeParseException{
```
forestmu: ```DateTimeParseException{``` don't you think there should be a space?

src/main/java/jamal/task/Event.java
```diff
@@ -0,0 +1,76 @@
...
+            this.end = LocalDateTime.parse(end);
+        } catch (DateTimeParseException e) {
+            throw e; //prevent creation of event object for snowballing errors
+        }
+
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (from: " + DateTime.formatDateTime(start) + " to: " + DateTime.formatDateTime(end) + ")";
```
forestmu: Don't you think this is a bit long?
I think it's better to break it into 2 lines

src/main/java/jamal/ui/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ *
+ * Initialise Scanner to detect inputs
+ * Print statements for Welcome, Exit and Linebreaks
+ *
+ */
+public class Ui {
+
+    final static String lineBreak = "____________________________________________________________\n";
```
forestmu: this is a constant so it isn't suppose to be in camelCase but rather all uppercase

src/main/java/jamal/ui/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+    public Ui() {
+        this.inputScanner = new Scanner(System.in);
+    }
+
+    /**
+     * Prints welcome statement
+     */
+    public void showWelcome() {
+        System.out.println(
+                lineBreak +
```
forestmu: it seems like the ```+``` should be in front instead of at the back

src/main/java/jamal/util/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+            } catch (ArrayIndexOutOfBoundsException e) {
+                throw new InvalidCommandException();
+            }
+        }
+
+        /// Event task
+        if (input.toLowerCase().startsWith("/event")) {
+            try {
+                String secondHalfString = input.split("\\s+", 2)[1];
+                Matcher eventInfo = eventInfoRegexPattern.matcher(secondHalfString); //refer to eventInfoRegexPattern at top of main, need to split into 3 groups for 3 params
```
forestmu: if the comment is a bit long, do you think it would be better to split into 2 lines? And then maybe comment at the top like how u did for ```/// Event task``` etc?

forestmu (commented):
I think it's overall good. 
There are header comments almost everywhere that's required.
Naming looks good (except the one part where you named the constant wrongly)
You might want to check the ordering of imports in all your command classes though.

src/main/java/jamal/command/Command.java
```diff
@@ -0,0 +1,31 @@
+package jamal.command;
+
+import jamal.task.*;
```
nghaowei: Perhaps you could consider importing specific libaries or packages instead of using a wildcard?

src/main/java/jamal/task/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+    }
+
+    /**
+     * Lists all tasks in tasklist
+     *
+     * @returns print statement of all tasks in tasklist
+     */
+    public void list() {
```
nghaowei: Perhaps you could consider using an action verb such as printList?

src/main/java/jamal/task/ToDo.java
nghaowei, line 1: Perhaps you could consider renaming this class to Todo (without the capital 'D'), so it aligns with common terminology?

nghaowei (commented):
LGTM! 👍

PR #269, opened 2025-08-26
[Bobbodi] iP

Bobbodi:
**Bobbodi is the chatbot to manage your tasks** 
- [ ] text-based
- [ ] colorful
- [ ] simple to use

All you need to do is
1. Add your tasks
2. let it manage your tasks for you 😉

To run the JAR file from the terminal, 

```
java -jar build/libs/CS2103T.jar
```


haojun-mah:
LGTM. Some nits to fix: Commits more than 72 chars, commits with periods at the end. Commits with starting char not caps.

src/main/java/Bobbodi.java
```diff
@@ -19,18 +19,20 @@ public static void main(String[] args) {
     public static final String BLUE = "\u001B[34m";
     public static final String GREEN = "\u001B[32m";
     public static final String PURPLE = "\u001B[35m";
+    public static final String SEPARATOR = BLUE + DASHES + RESET;
```
haojun-mah: Correct naming

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+
+    protected String description;
+    protected boolean isDone;
```
haojun-mah: good use of OOP concepts

src/main/java/Bobbodi.java
```diff
@@ -2,7 +2,10 @@
 import Tasks.*;
 import Resources.*;
```
haojun-mah: hmmmmm, *?

src/test/java/HelperTest.java
```diff
@@ -0,0 +1,19 @@
+import Resources.Helper;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class HelperTest {
+    @Test
```
haojun-mah, line 17: Would it be great if you describe what the test achieves here

PR #277, opened 2025-08-26
[Wei Xian] iP

weixianuni:
## Yappy :sunglasses: 

> "Outer order, inner calm" - Gretchen Rubin 

Yappy keeps your life ordered, so you can focus on what matters in life.

Yappy is,

* text-based
* easy-to-learn
* **EXTREMELY FAST**

To get started, all you need to do is,

1. download the app from [here](https://github.com/weixianuni/ip/releases/tag/A-Jar)
2. run the application 
3. input your tasks
4. let Yappy manage your tasks for you. 

Features:
- [X] Manage todos, deadlines and events
- [ ] Reminders (coming soon)

Here's the `main` method:
``` java
        new Yappy(storagePath).run();
}
```


src/main/java/yappy/backend/Storage.java
```diff
@@ -0,0 +1,88 @@
...
+        this.file = Paths.get(filePath).toFile();
+    }
+
+    /**
+     * Returns an ArrayList of the tasks that have been read from disk
+     *
+     * @return Tasks saved to disk from previous interaction
+     * @throws YappyException If format is incorrect
+     */
+    public ArrayList<Task> loadTask() throws YappyException{
```
goatmilkkk: The method name `loadTask()` suggests it loads a single task, but it actually returns a collection. Consider renaming to `loadTasks()` for clarity and consistency with the return type.

src/main/java/yappy/backend/Storage.java
```diff
@@ -0,0 +1,88 @@
...
+            while (scanner.hasNextLine()) {
+                String line = scanner.nextLine();
+                if (!line.trim().isEmpty()) {
+                    String[] params = line.split("\\|");
+                    switch (params[0]) {
+                    case "T":
+                        tasks.add(new ToDo(params[2], Boolean.parseBoolean(params[1])));
+                        break;
+                    case "E":
+                        tasks.add(new Event(params[2], Boolean.parseBoolean(params[1]), LocalDateTime.parse(params[3], FORMATTER), LocalDateTime.parse(params[4], FORMATTER)));
```
goatmilkkk: Consider breaking this line across multiple lines to comply with the 120-char limit in the Java coding standard.

src/main/java/yappy/ui/Yappy.java
```diff
@@ -0,0 +1,74 @@
...
+            ui.showWelcome();
+            ui.showLine();
+        }
+    }
+
+    /**
+     * Keeps the program alive and listens for user inputs.
+     */
+    public void run() {
+        boolean isExit = false;
```
goatmilkkk: I like how you named your `isExit` flag — the boolean name makes it easy to read and understand.

src/main/java/yappy/backend/TaskList.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Marks the task at position index in the ArrayList as uncompleted.
+     *
+     * @param index Index of the task that is to be marked as uncompleted.
+     * @return Task that was marked as uncompleted.
+     * @throws YappyException If index >= tasks.size() or index < 0.
+     */
+    public Task unmark(int index) throws YappyException {
+        if (tasks.isEmpty()) {
```
goatmilkkk, line 88: I think the `isEmpty()` check is redundant since the `IndexOutOfBoundsException` try-catch below already covers the empty-list case.

goatmilkkk (commented):
Great work overall! I like how you separated out your Commands into separate Java files for clarity and better organization :)

src/main/java/yappy/backend/Storage.java
weixianuni: Right good catch thanks.

src/main/java/yappy/backend/TaskList.java
weixianuni: Thank you for your comments, i think theyre great

src/main/java/yappy/task/Event.java
```diff
@@ -0,0 +1,66 @@
...
+ * a start date and an end date.
+ *
+ * <p>This class extends the <code>Task</code> class and overrides methods
+ * for file storage and string representation.</p>
+ */
+public class Event extends Task {
+
+    LocalDateTime from;
+    LocalDateTime to;
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
```
morpheuschoo: formatter seems to be a const. I think it would be good to declare it as "final" and change its name to FORMATTER.

morpheuschoo (commented):
In general, your naming conventions are very good. Just a few minor nit picks, but it should be okay 👍

PR #282, opened 2025-08-26
[jakekeech] iP

jakekeech:
# DUM-E

> "With great power comes great responsibility."

**DUM-E** is a *text-based* task manager, allowing you to create, search and save your tasks. 
~~Clunky~~ **Clean** and simple to use! 😊

**Features**
- Simple text-UI
- Persistent Storage
- Accurate search function

**How to Use**
1. Download the JAR [here](https://github.com/jakekeech/ip/releases/tag/A-Jar)
2. Run the file
3. Add tasks
4. Let DUM-E your tasks for you

**Upcoming Features**
- [x] Search Function
- [ ] GUI
- [ ] Greater Customisation

Here's the`main` method if you want to run it in java:
```java
public class Main {
    }
}
```


src/main/java/dume/Dume.java
XJie97, line 1: LGTM! I like that class files seem to be named clearly and descriptively, there is very little ambiguity as to which class does what from the naming convention. The class and package name matches the product name, it is easy to keep track of which files are related in the package.

src/main/java/dume/storage/Storage.java
XJie97, line 1: I like that the methods are named concisely and descriptively as verbs, this seems to adhere to Implementation → Code Quality → Naming → Basic → Use nouns for things and verbs for actions

src/main/java/dume/task/Task.java
XJie97, line 1: I only have a small nitpick here, and it is technically outside the scope of Code Quality: Naming.
Otherwise, the rest of it looks good to me.

Would it be better to rename the booleans as 'isDone' instead of 'done', and 'isMatch' instead of 'matches' as per the coding standards reference page?

w/ ref:
https://se-education.org/guides/conventions/java/intermediate.html
Section: Boolean variables/methods should be named to sound like booleans

image

src/main/java/dume/task/Task.java
```diff
@@ -0,0 +1,67 @@
+package dume.task;
+
+/**
+ * Represents a generic task with a description and completion status.
+ */
+public abstract class Task {
+    protected final String details;
+    protected boolean done;
```
XJie97: Would it be better to rename this as isDone?

src/main/java/dume/task/Task.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Checks if this task's details contain the given query.
+     * The match is case-insensitive.
+     *
+     * @param query the keyword to search for
+     * @return true if details contain the query, false otherwise
+     */
+    public boolean matches(String query) {
```
XJie97, line 59: Would it be better to rename this as isMatch instead? To make it clear to a reader that this is a boolean when it is called somewhere in the code.

XJie97 (commented):
Overall, the PR seems to adhere to naming conventions, outside of a few boolean variables. I think that this coding style emphasizes conciseness and makes life easier for the programmer, without sacrificing too much in readability.

LGTM, I really couldn't find many problems with naming convention in here, so there are not many comments from me, good job!

ZavierCSJ (commented):
“LGTM. Honestly, I don’t have much to comment on—other than what @XJie97 mentioned, everything else looks good!”

PR #283, opened 2025-08-26
[trekkie3] iP

trekkie3:
# Iris

> Hello! I'm **Iris**.
> What can I do for you?

## About Iris

- ✅ Siri backwards (Don't tell Apple)
- ✅ Unironically more useful than Siri
- ✅ Text-based for now. More features incoming.

## Features

1. Easy to use.
2. Automatically saves and loads to disk.
3. Completely free.

## Using Iris

To use, simply click the run button or call `java -jar` with the compiled JAR file:

```sh
java -jar build/libs/Iris.jar
```

## To Be Implemented

- [ ] Implement **_GUI_**
- [ ] Easier task inputs
- [ ] Add more task types (_recurring tasks_)

## Contribute

Want to contribute? [Simply open a pull request here.](https://github.com/trekkie3/ip)


src/main/java/Iris.java
```diff
@@ -0,0 +1,179 @@
+import task.Deadline;
+import task.Event;
+import task.Task;
+import task.Todo;
+
+import java.io.*;
```
cyfake: Should the imported classes be listed explicitly instead of using ```*```? (as per the [guidelines](https://se-education.org/guides/conventions/java/intermediate.html#package-and-import-statements))

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Generates a Deadline task from the given argument string.
+     *
+     * @param argument string containing details
+     * @return Deadline task
+     * @throws TaskException on invalid argument
+     */
+    public static Deadline generateDeadline(String argument) throws TaskException {
```
cyfake: Good naming as it is a verb and aptly describes what it is generating (a ```Deadline```)

src/main/java/CommandType.java
```diff
@@ -0,0 +1,12 @@
...
+    ADD_DEADLINE,
+    ADD_EVENT,
+    FIND,
+    DELETE,
+    MARK,
+    UNMARK,
+    LIST,
+    BYE,
+    INVALID
+}
```
cyfake: Good thinking to use enums for commands 👍

src/main/java/Iris.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Iris {
+    static final String SEPARATOR = "-------------------------------------";
+
+    static void preamble() {
```
cyfake: Perhaps it would be better to rename these methods as verbs, such as ```printPreamble()``` or ```sayFarewell()``` as per the [naming guidelines](https://nus-cs2103-ay2526s1.github.io/website/schedule/week4/topics.html#w4-6-code-quality-naming)?

src/main/java/Iris.java
trekkie3: Thanks for the heads up, I have fixed the issue.

src/main/java/Iris.java
trekkie3: You're definitely right, I should fix that. Thanks!

src/main/java/Iris.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public static void main(String[] args) {
+        Scanner reader = new Scanner(System.in);
+        String filePath = "data.txt";
+        List<Task> taskList = load(filePath);
+
+        preamble();
+
+        MAIN:
+        while (true) {
```
balkinaveen: I like the implementation here and the syntax looks good to me. Furthermore, the handling of exception is logical, printing out the usage hint only in the event of an invalid input. Good job!

src/main/java/task/Event.java
```diff
@@ -0,0 +1,89 @@
...
+package task;
+
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task {
+    final LocalDate from;
+    final LocalDate to;
+    private static final DateTimeFormatter PARSER = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
balkinaveen: Good idea to fix the format as a constant. Perhaps it might be a good idea to consider creating a DateTimeFormatterUtil class to handle the parsing of String, date values and returning the Date in the specified patterns. This idea was suggested to me and I thought it might be applicable in yours as well. This would prevent code duplication.

src/main/java/task/TaskExceptionType.java
```diff
@@ -0,0 +1,8 @@
+package task;
+
```
balkinaveen: Intuitive implementation of exception handling. LGTM!

src/test/java/task/DeadlineTest.java
```diff
@@ -0,0 +1,24 @@
+package task;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
balkinaveen: I’m not sure if we can do a * import here. I believe it would be better practice to specify the list the imports.

trekkie3: Thank you for pointing this out. You are currently viewing older code and this mistake has already been fixed in the most current version. If you can find any more code style errors, please let me know, thank you!

PR #285, opened 2025-08-26
[makis4n] iP

makis4n:
# LeBron
> “I'm going to use all my tools, my God-given ability, and make the best life I can with it.” ― LeBron James [(source)](https://www.goodreads.com/author/quotes/2933701.LeBron_James)

LeBron is at good at remembering your tasks as he is at basketball 🏀 , he is:
- Text-based
- As easy to use as a layup
- ~~FAST~~ SUPER FAST to use

All you need to do is:
1. Download
2. Double-click it
3. Add your tasks
4. Let him cook ⛹️‍♂️ 

And its all **FREE**!

Features:
- [x] Managing tasks
- [ ] GUI (coming soon)
---
If you want to be as good at Java as LeBron, you can use it to practice Java too. Here's the `main` method:
```ruby
        new LeBron("data/tasks.txt").run();
}
```


src/main/java/lebron/command/FindCommand.java
```diff
@@ -0,0 +1,43 @@
+package lebron.command;
+
+import lebron.task.*;
```
katrinaykt: perhaps the imported classes should be listed explicitly instead of using a wildcard import?

src/main/java/lebron/command/UnmarkCommand.java
```diff
@@ -0,0 +1,37 @@
...
+import lebron.main.Storage;
+import lebron.main.Ui;
+import lebron.task.Task;
+import lebron.task.TaskList;
+
+public class UnmarkCommand extends Command {
+    private String arguments;
+
+    /* Constructor for UnmarkCommand class
+     */
```
katrinaykt, line 18: For Javadoc comments, it's best to use `/** ... */` instead of `/* ... */` so that documentation tools can recognise and process the comments correctly! (same for the other javadoc comments)

katrinaykt (commented):
Looks good overall! I didn't notice much issues with the code quality other than the minor issues mentioned :)

src/main/java/lebron/command/FindCommand.java
llamawithhats: I second what @katrinaykt has mentioned about wildcard imports.

src/main/java/lebron/command/AddCommand.java
```diff
@@ -0,0 +1,118 @@
...
+import lebron.exception.LeBronException;
+import lebron.main.Storage;
+import lebron.main.Ui;
+import lebron.task.Deadline;
+import lebron.task.Event;
+import lebron.task.Task;
+import lebron.task.TaskList;
+import lebron.task.ToDo;
+
+public class AddCommand extends Command {
```
llamawithhats, line 19: Perhaps class names should be in the form of nouns instead of verbs? This applies to your other Command subclasses.

src/main/java/lebron/command/AddCommand.java
```diff
@@ -0,0 +1,118 @@
...
+    private String commandWord;
+    private String arguments;
+
+    /* Constructor for AddCommand class.
+     *
+     * @param arguments The arguments provided with the add command.
+     */
```
llamawithhats: Similar to what the other reviewer mentioned:
perhaps your header comments should be in the form of 
```
/**
 * Returns lateral location of the specified position.
 * If the position is unset, NaN is returned.
 *
 * @param x X coordinate of position.
 * @param y Y coordinate of position.
 * @param zone Zone of position.
 * @return Lateral location.
 * @throws IllegalArgumentException If zone is <= 0.
 */ 
```
where `/**` is on a separate line.

llamawithhats (commented):
Looks good 👍 
Found some coding standard violations that you can work on.

PR #287, opened 2025-08-26
[cyfake] iP

cyfake:
Baymax is a friendly, text-based chatbot that helps you remember things you need to do, events and deadlines. It's,
- text-based,
- easy to learn
- ~FAST~ _SUPER FAST_ to use
- always making sure you are satisfied with your care

All you need to do is,
1. download it from [here](https://github.com/cyfake/ip/releases)
2. run the command `java -jar "baymax.jar"` in the same folder as the jar file
3. add your tasks
4. let Baymax manage your tasks for you! 😸 

And it is **FREE**!

Existing features:
- [x] Managing tasks
- [x] Managing deadlines 
- [x] Managing events
- [ ] Reminders (coming soon)
---
```java
        new Baymax("./data/Baymax.txt").run();
    }
```


src/main/java/baymax/parser/Parser.java
```diff
@@ -0,0 +1,110 @@
...
+public class Parser {
+
+    /**
+     * Parses the index argument from a command input.
+     *
+     * @param arr The array of command tokens.
+     * @return The task index.
+     * @throws NumberFormatException If the index cannot be parsed as an integer.
+     *
+     */
```
haorui2004, line 31: I like how descriptive your javadocs are, they have really helped me in understanding all the code efficiently

src/main/java/baymax/task/Task.java
```diff
@@ -0,0 +1,130 @@
...
+    public String toString() {
+        return String.format("[%s] %s", this.getStatus(), this.description);
+    }
+
+    /**
+     * Marks this task as completed and returns a confirmation message.
+     *
+     * @return A confirmation string including the task.
+     */
+    public String mark() {
```
haorui2004, line 36: Just suggesting an alternative here: I think naming the method markDone or markComplete can be more straightforward and easier for people to see. But your current name is also ok!

src/main/java/Baymax.java
```diff
@@ -0,0 +1,55 @@
...
+            try {
+                storage.save(tasks);
+            } catch (IOException e) {
+                ui.printMsg("Error saving tasks: " + e.getMessage());
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        new Baymax("./data/Baymax.txt").run();
```
haorui2004: I guess the file path can just be saved as a constant since we are not changing it later. (Just for easy reference)

src/main/java/baymax/command/UpdateCommand.java
```diff
@@ -0,0 +1,50 @@
...
+            ui.printMsg(tasks.mark(index));
+            break;
+        case "unmark":
+            ui.printMsg(tasks.unmark(index));
+            break;
+        case "delete":
+            ui.printMsg(tasks.delete(index));
+            break;
+        }
+    }
```
haorui2004, line 47: I like how you group them together! This is a smart approach

src/main/java/baymax/task/Task.java
```diff
@@ -0,0 +1,130 @@
...
+        this.isDone = isDone;
+        this.description = description;
+    }
+
+    /**
+     * Returns the completion marker used in string renderings.
+     *
+     * @return {@code "X"} if done, otherwise a single space {@code " "}.
+     */
+    public String getStatus() {
```
haorui2004, line 22: The method looks good, just a bit nitpicking here, perhaps we could name it getStatusIcon to make it easier to read? Since anyways we are using this to reflect the isDone status visually on outputs

src/main/java/baymax/exception/BaymaxException.java
```diff
@@ -0,0 +1,50 @@
+package baymax.exception;
+
+public class BaymaxException extends Exception {
```
haorui2004, line 3: I like the design of the exception class here! It is really nice that you separated all these cases into individual classes, this makes things very clear when encountering exceptions

haorui2004 (commented):
Overall your code is complying with the coding standards well, good job! The comments are just about minor suggestions.

PR #317, opened 2025-08-27
[Kulkarni Venugopal Vasant] iP

Vorld:
# Chuck Task Manager 🤖

A simple task management chatbot built in Java.

> "In the book of life, the answers aren't in the back." - Charlie Brown

## Features
- Add and manage tasks
- Mark tasks as done
- Find tasks by keyword
- Save tasks automatically

## Getting Started

1. Clone the repository
2. Compile the Java files
3. Run the main class
4. Start chatting with Chuck!

### Example Usage

Here is a sample `main` method.

```java
public class Chuck {
        TaskList tasks = new TaskList();
        tasks.addTask("Buy groceries");
        System.out.println("Task added!");
    }
}
```

## Todo
- [x] Basic task management
- [x] File I/O operations
- [ ] GUI implementation
- [ ] Advanced search features

## Download

**Download** [*now!*](https://github.com/Vorld/ip/releases/tag/A-Jar).


data/chuck.txt
```diff
@@ -0,0 +1 @@
+D | false | submit report | 2023-12-01 23:59
```
its-me-orion: Not sure if this was intentional, but I believed that this is part of the app local storage? If that is the case, may I suggest that this should be included in `.gitignore`?

src/main/java/chuck/task/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+import chuck.command.Parser;
+import java.time.LocalDateTime;
+
+/**
+ * Represents a task with a deadline (due by specific date/time).
+ * Extends the basic Task functionality with a deadline date and time.
+ */
+public class Deadline extends Task {
+    public static final String SHORT_HAND = "D";
+    protected LocalDateTime by;
```
its-me-orion, line 17: Do you think if sticking by this variable name "by" better or would a more descriptive variable like "deadline" or "dueDate" be better?

In addition to this, the variable "SHORT_HAND" is not that intuitive to me, maybe a more understood name could be "LABEL" or "TASKTYPE"?

src/main/java/chuck/task/Event.java
```diff
@@ -0,0 +1,63 @@
...
+import chuck.command.Parser;
+
+import java.time.LocalDateTime;
+
+/**
+ * Represents an event task with start and end times.
+ * Extends the basic Task functionality with from and to date/time fields.
+ */
+public class Event extends Task{
+    public static final String SHORT_HAND = "E";
```
its-me-orion, line 12: I find this variable name not that intuitive, maybe a more intuitive name could be "LABEL" or  "TASK_TYPE"?

src/main/java/chuck/task/Event.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+    /**
+     * Returns formatted string with [E] prefix and event time range.
+     *
+     * @return string representation with event type indicator and time range
+     */
+    @Override
+    public String toString() {
+        return String.format("[%s]%s (from: %s to: %s)", SHORT_HAND, super.toString(), Parser.formatDateTime(this.from), Parser.formatDateTime(this.to));
```
its-me-orion: I believe this can be broken down into multiple lines instead of a single line.

src/main/java/chuck/ui/Ui.java
```diff
@@ -0,0 +1,48 @@
...
+        return scanner.nextLine();
+    }
+
+    /**
+     * Displays a message to user with borders.
+     *
+     * @param message the message to display
+     */
+    public void showMessage(String message) {
+        System.out.println("____________________________________________________________");
```
its-me-orion: Personally for mine I have a global variable for line at the top of this class, that way if I need to change the UI of the line I can do it in one shot, while there is nothing wrong with this at the moment, I think is something you could consider.

its-me-orion (commented):
Overall, the code is neatly distributed and structured, making it easy to follow. Just some nits to pick. 👍

PR #354, opened 2025-08-27
[zayyankece] iP

zayyankece:
### JOKO TASK MANAGER
-------------------------------------------------------------------

> “I love deadlines. I like the whooshing sound they make as they fly by.” – Douglas Adams

### Key Features 
-----------------------------------------------------------------
**Joko** helps you keep track of your tasks. It is:

- Text-based
- Simple to use
- Fast ⚡

### Getting Started
-----------------------------------------------------------------
Getting started is easy:

1. **Download** it from [here](https://github.com/zayyankece/ip)
2. **Run** the application
3. **Add** your tasks
4. **Let it manage** them for you 😉

And yes, it is free!

### Some Features
-----------------------------------------------------------------
Features:

- [X]  Managing tasks
- [X]  Managing deadlines
- [ ] Reminders (coming soon)

### Code Structure
-----------------------------------------------------------------
For Java programmers, you can **practice coding** too. Here's the main method:
```
public class Joko {
        Application.launch(JokoApp.class, args);
    }
}
```

Some additional points:

- You can mark tasks as done using `mark <task number>`
- You can unmark tasks using `unmark <task number>`
- Say no to Old CLI commands :P


src/main/java/Joko.java
```diff
@@ -0,0 +1,263 @@
...
+                System.out.println("sorry i could not understand your command :(");
+                System.out.println("____________________________________________________________");
+            }
+        }
+
+        sc.close();
+    }
+
+    static class Task {
+        protected final String desc;
```
abhatia2003: Perhaps a better naming style could be ```DESCRIPTION```, since ```desc``` is a constant, to follow the Java Coding Standard.

src/main/java/Joko.java
```diff
@@ -0,0 +1,263 @@
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Joko {
```
abhatia2003: You can consider adding a JavaDoc header for public methods and classes, to make the code more informative for users that may not be accustomed to Java.

src/main/java/Joko.java
```diff
@@ -0,0 +1,263 @@
...
+                        }
+                        newTask = new ToDo(desc);
+                    } catch (Exception e) {
+                        System.out.println("Error adding Todo: " + e.getMessage());
+                    }
+                } else if (command.equals("deadline")) {
+                    try {
+                        String[] parts = input.substring(9).split(" /by", 2);
+                        if (parts.length < 2 || parts[0].trim().isEmpty() || parts[1].trim().isEmpty()) {
+                            throw new IllegalArgumentException("Deadline must have a description and " +
```
abhatia2003: Just a small suggestion, having the ```+``` on the next line could be better as then, it would be in line with the Java Coding Standard. For instance, in this case, it would be as follows: 
```throw new IllegalArgumentException("Deadline must have a description and "```
```+ "a /by date/time (dd/MM/yyyy HHmm)");```

Lines 176, 204 and 205 can also be modified to reflect this change and more closely follow the Java Coding Standard.

File: src/main/java/Joko.java [hunk 4]
```diff
@@ -0,0 +1,263 @@
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
```
abhatia2003: Can consider swapping line 6 with line 7 to ensure that the imports are in lexicographical order and hence follow the Java Coding Standard.

abhatia2003 (commented):
Code mostly follows the Java Coding Standard, which is great to see! Just need to fix some small details and it should be fully compliant with the Coding Standard, nicely done!

src/main/java/Joko.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return getStatus() + desc;
+        }
+    }
+
+    static class Deadline extends Task {
+        protected LocalDateTime by;
```
keitht8y:
```suggestion
        protected LocalDateTime endDate
```

src/main/java/Storage.java
```diff
@@ -0,0 +1,67 @@
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    private final String filename;
```
keitht8y:
```suggestion
    private final String fileName
```

src/main/java/Ui.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class Ui {
+    private final Scanner sc;
+
+    public Ui() {
+        this.sc = new Scanner(System.in);
+    }
+
+    public void showWelcome(ArrayList<Joko.Task> tasks) {
+        System.out.println("____________________________________________________________");
```
keitht8y: Perhaps saving the message divisors as a reusable constant to be printed might help to save some code space 💭?

keitht8y (commented):
LGTM 👍. Just a few nits to fix.

PR #369, opened 2025-08-27
[Juink4i] ip

Juink4i:
# NaiLongBot
> Ni Hao

NaiLongBot frees your mind of having to remember things you need to do. It's,

text-based
easy to learn
FAST SUPER FAST to use
All you need to do is,

download it from [here](https://github.com/Juink4i/ip).
double-click it.
add your tasks.
let it manage your tasks for you 😉
And it is FREE!

Features:

- [x] Managing tasks
- [x] Managing deadlines (coming soon)
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Jkbot/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+import Jkbot.utils.Storage;
+import Jkbot.utils.TaskList;
+import Jkbot.utils.Ui;
+import Jkbot.exception.JkBotException;
+
+/**
+ * Command to add a new deadline task.
+ */
+public class AddDeadlineCommand implements Command {
```
Vrospix: Perhaps a noun class name would be more suitable?
```suggestion
public class DeadlineCommand implements Command {
```

src/main/java/Jkbot/exception/InvalidFormatException.java
```diff
@@ -0,0 +1,7 @@
+package Jkbot.exception;
+
+public class InvalidFormatException extends JkBotException{
```
Vrospix:
```suggestion
public class InvalidFormatException extends JkBotException {
```

src/main/java/Jkbot/exception/JkBotException.java
```diff
@@ -0,0 +1,8 @@
+package Jkbot.exception;
+
+public class JkBotException extends Exception{
```
Vrospix: Perhaps you can start tidying up the codes as per [coding standards](https://se-education.org/guides/conventions/java/intermediate.html) as I noticed this issue in several other files :)
```suggestion
public class JkBotException extends Exception {
```

src/main/java/Jkbot/task/Deadline.java
```diff
@@ -0,0 +1,38 @@
+package Jkbot.task;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
```
Vrospix: I also noticed some classes doesn't have the required JavaDoc yet. Perhaps you can start adding it as well.

src/main/java/Jkbot/utils/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+            return new ShowCommand(arguments);
+
+        case "find":
+            validateNotEmpty(arguments, "find [key words]");
+            return new FindCommand(arguments);
+
+        default:
+            return new UnknownCommand();
+        }
+    }
```
Vrospix: I like your way of using cases here 💯
Here are some thoughts :
_How would it be if you implement enumerations for case by case scenario?_

src/main/java/Jkbot/utils/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+     * Parses a single line from the data file into a Task object.
+     * Expected format: "T | 0 | description" for Todo,
+     *                 "D | 1 | description | byDate" for Deadline,
+     *                 "E | 0 | description | fromTime | toTime" for Event.
+     *
+     * @param input the file line to parse, formatted with " | " separators
+     * @return the parsed Task object, or null if input format is invalid
+     * @throws ArrayIndexOutOfBoundsException if input doesn't have enough parts
+     * @throws IllegalArgumentException if task type is unrecognized
+     */
```
Vrospix: I really like your implementation of JavaDoc here. Truly detailed 👍🏻

src/test/java/Jkbot/command/AddDeadlineCommandTester.java
```diff
@@ -0,0 +1,57 @@
...
+        assertTrue(output.contains("Submit report"));
+        assertTrue(output.contains("Now you have 1 tasks in the list"));
+    }
+
+    @Test
+    void isExitReturnsFalse() {
+        AddDeadlineCommand command = new AddDeadlineCommand("Submit report /by 28/8/2025 2359");
+        assertFalse(command.isExit());
+    }
+}
```
Vrospix: Great to see the amount of test files you have here! Good job 👍🏻

Vrospix (commented):
Overall looks really solid, great use of OOP, and nice amounts of testers. Styling can be slightly improved. Otherwise, keep it up! 👍🏻

src/main/java/Jkbot/command/AddDeadlineCommand.java
yoyongxuan: I believe the name AddDeadlineCommand is meant to be a noun as it refers to the command to "add a deadline", rather than the action of adding a "deadline command".

However, changing the name to DeadlineCommand would be less confusing and you should consider such a change.

src/main/java/Jkbot/command/Command.java
```diff
@@ -0,0 +1,28 @@
...
+     * @throws JkBotException if command execution fails
+     */
+    void execute(TaskList tasks, Ui ui, Storage storage) throws JkBotException;
+
+    /**
+     * Indicates whether this command should exit the application.
+     *
+     * @return true if this is an exit command, false otherwise
+     */
+    boolean isExit();
```
yoyongxuan: ### Should the name of isExit be a verb instead?

You have named this method according to the naming standard for booleans, which may be appropriate given that this method returns a boolean and naming it as one does describe its function appropriately. 
However, as this is still technically a method, consider naming it a verb instead such as:
```suggestion
    boolean checkExit();
```

src/main/java/Jkbot/utils/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+public class Parser {
+
+    public static Command parse(String input) throws JkBotException {
+        if (input.isEmpty()) {
+            throw new EmptyDescriptionException("Empty command");
+        }
+
+        String[] parts = input.split(" ", 2);
+        String command = parts[0].toLowerCase();
+        String arguments = parts.length > 1 ? parts[1] : "";
```
yoyongxuan: ### Should arguments be singular?

As you have named arguments in plural form, this suggests that arguments is an array or some kind of multi-valued variable despite being a String. 
Consider changing the name to reflect this:
```suggestion
        String argumentStr = parts.length > 1 ? parts[1] : "";
```

src/main/java/Jkbot/utils/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+import Jkbot.task.Event;
+
+
+/**
+ * Manages the storage and retrieval of tasks from file system.
+ * Handles both loading tasks on startup and saving tasks on exit.
+ */
+public class Storage {
+    private static final String FILE_PATH = "data/saved_tasks.txt";
+    private static final String DIRECTORY_PATH = "data";
```
yoyongxuan: ### Should constants have a common prefix?

The Java Coding Standard states that

> Associated constants should have a common prefix.

while your constants have a common suffix (PATH) instead.
Consider moving the word PATH to the front of each name to make it a prefix instead.

src/main/java/Jkbot/utils/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Loads the task from saved location on start up.
+     *
+     * @return the ArrayList of Tasks that are stored.
+     * @throws IOException when file is not created.
+     */
+    public ArrayList<Task> loadTasks() {
+        ArrayList<Task> memory = new ArrayList<>();
```
yoyongxuan: ### Should memory be plural?

As memory is an array of tasks, consider giving it a name in plural form instead:
```suggestion
        ArrayList<Task> loadedTasks = new ArrayList<>();
```

src/main/java/Jkbot/utils/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            Path path = Paths.get(FILE_PATH);
+            if (!Files.exists(path)) {
+                return memory; // No file to load
+            }
+            List<String> lines = Files.readAllLines((path));
+            int corruptedLines = 0;
```
yoyongxuan: ### Should corruptedLines be singular?
As corruptedLines is an integer representing the number of corrupted lines, rather than an array of corrupted lines, consider renaming it to better reflect its contents and function:

```suggestion
            int corruptedLineCount = 0;
```

src/main/java/Jkbot/utils/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+                    break;
+                case "D":
+                    task = new Deadline(desc, parts[3]);
+                    break;
+                case "E":
+                    task = new Event(desc, parts[3], parts[4]);
+                    break;
+                default:
+                    return null;
+            }
```
yoyongxuan: ### Should `case ` clauses be indented?

According to the Java Coding standard, there is no indentation for case clauses in switch statements.
Consider updating your code to reflect this:
```suggestion
            switch(type) {
            case "T":
                task = new Todo(desc);
                break;
            case "D":
                task = new Deadline(desc, parts[3]);
                break;
            case "E":
                task = new Event(desc, parts[3], parts[4]);
                break;
            default:
                return null;
            }
```

src/main/java/Jkbot/utils/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Handles all user interface interactions.
+ * Responsible for displaying messages and reading user input.
+ */
+public class Ui {
+    private final Scanner scanner;
+    public static String LINE = "_____________________________________________________________\n";
+    public static String OPENING = LINE + "Hello! I'm jkbot\nWhat can I do for you?\n" + LINE;
+    public static String CLOSING = LINE + "Bye. Hope to see you again soon!" + "\n" + LINE;
```
yoyongxuan: ### Should these constants be more descriptive?

Consider changing their names to better describe their contents and purpose:
```suggestion
    public static String OUTPUT_LINE = "_____________________________________________________________\n";
    public static String MESSAGE_OPENING = LINE + "Hello! I'm jkbot\nWhat can I do for you?\n" + LINE;
    public static String MESSAGE_CLOSING = LINE + "Bye. Hope to see you again soon!" + "\n" + LINE;
```

yoyongxuan (commented):
Overall tidy and easily readable code.  Mostly follows Java Naming Standards with a few minor exceptions.

PR #371, opened 2025-08-27
[yjjr0] iP

yjjr0:


src/main/java/yapbot/parser/Parser.java
```diff
@@ -0,0 +1,269 @@
...
+            UI.invalidFile();
+        }
+    }
+
+    /**
+     * Processes the CLI input by the specified command
+     *
+     * @param task the CLI input from the user
+     */
+    public static void setTask(String task) {
```
jvvvstin: Could a better variable name be used here instead of `task`? Maybe `input`?

src/main/java/yapbot/parser/Parser.java
```diff
@@ -0,0 +1,269 @@
...
+     * @return true if the task is marked, else false
+     */
+    public static boolean mark(String input) {
+        try {
+            int index = getTaskNumber(input);
+            Task task = TaskList.getTask(index);
+            task.mark();
+            System.out.println("     Nice! I've marked this task as done:\n" +
+                    "        " + task + "\n" +
+                    "____________________________________________________________");
```
jvvvstin: Perhaps the displaying of output could be printed by a different class like Ui? Could the Parser class return an output meaningful to the Ui class to be printed?

src/main/java/yapbot/parser/ParserTest.java
jvvvstin, line 1: Perhaps this test file should be nested in src/test/java/yapbot/parser/ instead of its current directory

src/main/java/yapbot/taskmanager/DateTime.java
```diff
@@ -0,0 +1,61 @@
+package yapbot.taskmanager;
+
+import yapbot.ui.UI;
+
+import java.time.LocalDate;
+import java.time.Month;
+import java.time.format.DateTimeParseException;
+
```
jvvvstin, line 8: I like that you created a utility class to handle date time formatting. Perhaps you could add a header comment to this class so that others would be able to tell what this class does.

src/main/java/yapbot/taskmanager/EventTask.java
```diff
@@ -0,0 +1,21 @@
+package yapbot.taskmanager;
+
+public class EventTask extends Task {
+    private String startDateTime;
+    private String endDateTime;
```
jvvvstin: Any reason why these fields are of String type when the deadline field in DeadlineTask class is of LocalDate type?

jvvvstin (commented):
Overall, I found your code easy to read for the most part. I noted a few coding standard violations. Also, some of the classes are getting quite long, and perhaps you could consider splitting them into smaller classes if it is logical to do so.

src/main/java/yapbot/parser/Parser.java
balkinaveen: yep, i was thinking of suggesting this as well. Perhaps printing by a different class like Ui can help to reduce code duplication?

src/main/java/yapbot/ui/UI.java
```diff
@@ -0,0 +1,148 @@
+package yapbot.ui;
+
+public class UI
+{
```
balkinaveen: LGTM, I like how you have segregated and formatted the different UI messages.

PR #376, opened 2025-08-27
[tofuuu67] iP

tofuuu67:
### Heimerdinger Chatbot

> "Why do people say 'you can't trust atoms'? Because they 'make up everything'! Hehe!" - Heimerdinger, League of Legends

Heimerdinger is your go to chatbot because:

- It never betrays you ❤️ 
- It values precision 🔢 
- It has ~~good~~ perfect memory 🧠 

All you need to do is:

1. Download it from [here](https://github.com/tofuuu67/ip#)
2. Run the Heimerdinger.java class in IntelliJ
3. Add your tasks
4. Heimerdinger remembers everything for you!

And it is **FREE**!!

Here are our core features:

- [x] Adding of tasks
- [ ] Saving to disk (coming soon)
- [ ] Further UI personalisation (coming soon)

---
Here's the `main` method:
```ruby
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/heimerdinger/Storage.java [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Converts a Task into a string representation
+     *
+     * @param task the task to convert into string format
+     * @return string representation of the task
+     */
+    public String write(Task task) {
+        String type = task.getIcon();
+        String done = task.getStatusIcon().equals("X") ? "1" : "0";
```
Rah-Rah-Mitra, line 178: The variable names `type` and `done` are a bit brief. For greater clarity, what do you think about using slightly more explanatory names, such as `taskType` and `statusIndicator`?

src/main/java/heimerdinger/command/FindCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws HeimerdingerException {
+        if (matchingWord.isEmpty()) {
+            throw new HeimerdingerException("I can't read your mind great scientist. Let me know what you are trying to find!");
+        } else {
+            String[] split = matchingWord.split(" ");
+            if (split.length > 1) {
+                throw new HeimerdingerException("You should only enter 1 keyword for me to look for!");
+            }
+        }
+        TaskList displayList = new TaskList();
```
Rah-Rah-Mitra, line 57: I'm curious about the name `displayList`. Since this list contains the tasks that match the search, would a name like `foundTasks` or `matchingTasks` be a more accurate reflection of its purpose?

src/main/java/heimerdinger/task/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    protected String icon = "D";
+    protected LocalDate deadline;
+
+    public Deadline(String description, LocalDate deadline) {
+        super(description);
+        this.deadline = deadline;
```
Rah-Rah-Mitra: The field is named `deadline`, which is the same as the class name. This is quite common, but I wonder if a more specific name like `dueDate` or `byDate` might prevent any potential ambiguity, especially in more complex code. What are your thoughts on this?

data/heimerdinger.txt
Rah-Rah-Mitra, line 1: I see these files under the data directory are used for storing the task list data. It seems like the application updates them during runtime.

To prevent accidental commits of test data and avoid potential merge conflicts when multiple people run the application locally, have you considered adding the `data/` directory to your `.gitignore` file?

This would ensure that each developer's local data store doesn't get checked into version control, and the application could just create these files if they're missing on a fresh clone.

q
Rah-Rah-Mitra, line 1: I noticed these two new files, `q` and `tatus`, in the pull request. They appear to contain output from `git` commands (like `git log` or `git status`). Was it intentional to add these to the repository?

If they were added by accident, it might be a good idea to remove them and add their filenames to `.gitignore` to prevent them from being committed in the future.

Rah-Rah-Mitra (commented):
I like the thematic naming around "Heimerdinger," which adds a nice touch of personality to the project! The package structure (`command`, `task`) is also very clear and helps organize the code logically. The general commands and their outputs were very funny and I felt that personal touch was impressive (I will take notes for my own project). Generally, I had fun reviewing this pull request because of its uniqueness.

I also noticed a couple of files named `q` and `tatus` in the root directory. They look like output from a `git` command. Was it intentional to commit these? Keeping the repository clean of generated or temporary files can make it easier to navigate.

Here are a few specific suggestions regarding naming that you might consider.

data/heimerdinger.txt
ren0716, line 1: perhaps this could be added to .gitignore together with tasks.txt

ren0716 (commented):
overall i really like the good packaging and modularization of the code. I can understand the flow of the code really easily, well done! There are some minor coding standard violation as per other commenters but otherwise lgtm!

PR #395, opened 2025-08-27
Add an introduction to the README.md

chcischc:
# Capybara 🐹

> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://gettingthingsdone.com))

Capybara is a simple text-based task manager. It is,

- **text-based**
- *easy to learn*
- ~~boring~~ **FAST, SUPER FAST** 😉

All you need to do is:

1. Download it from [Releases](https://github.com/your-username/ip/releases).
2. Run it with `java -jar Capybara.jar`.
3. Add your tasks.
4. Let Capybara manage them for you 🚀

---

### Features

- [x] Manage tasks
- [x] Deadlines
- [x] Events
- [ ] Reminders (coming soon)

Here’s a snippet of the main method:

```java
public class Capybara {
        new Capybara("./Data/taskStorage.txt").run();
    }
}

### CODE REVIEW

File: src/main/java/capybara/Capybara.java [hunk 1]
```diff
@@ -0,0 +1,67 @@
+package capybara;
+
+import capybara.command.Command;
+
+/**
+ * Entry point of the Capybara application.
+ * Coordinates initialization of UI, storage, and task list, and
+ * runs the main command loop until the user exits.
+ */
+public class Capybara {
```
robinlogic, line 13: Great name! HAHA

src/main/java/capybara/Deadline.java
```diff
@@ -0,0 +1,23 @@
...
+public class Deadline extends Task {
+    private final LocalDateTime by;
+
+    public Deadline(String name, LocalDateTime by) {
+        super(name);
+        this.by = by;
+    }
+
+    @Override
+    public String toFileString() {
```
robinlogic, line 35: This is an encoding action which, according to OOP should be handled by an `encoder` in `Storage` class. BUT, I did the exact same thing as you to keep the implementation clean and abstracted. Just FYI

src/main/java/capybara/EmptyTimeException.java
```diff
@@ -0,0 +1,24 @@
+package capybara;
+
+public class EmptyTimeException extends CapyException {
+    public EmptyTimeException(String kind) {
+        super(buildMessage(kind));
+    }
+
+    private static String buildMessage(String kind) {
```
robinlogic, line 29: HAHAHA love this!

src/main/java/capybara/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+        String head = parts[0].toLowerCase();
+        String args;
+
+        if (parts.length > 1) {
+            args = parts[1].trim();
+        } else {
+            args = "";
+        }
+
+        if (head.equals("bye")) {
```
robinlogic: You could consider using Case-Switch with Enums for your Commands. It results in cleaner code and faster execution.

src/main/java/capybara/Storage.java
```diff
@@ -0,0 +1,140 @@
...
+                    LocalDateTime by;
+                    try {
+                        // Try full date-time first
+                        by = LocalDateTime.parse(raw, DATETIME_FMT);
+                    } catch (java.time.format.DateTimeParseException e1) {
+                        try {
+                            // Fallback: date-only → midnight
+                            by = LocalDate.parse(raw, DATE_FMT).atStartOfDay();
+                        } catch (java.time.format.DateTimeParseException e2) {
+                            System.out.println("capybara.Capybara can’t read that date 🐹🍊. Try 'yyyy-MM-dd' or 'yyyy-MM-dd HH:mm'.");
```
robinlogic: I actually didn't know you can use emojis... Aight, Imma steal this idea hehe

robinlogic: On a side note, I think some of these lines are more than 120 characters long. (94, 114 and 125)

src/main/java/capybara/TaskList.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Manages a dynamic list of tasks.
+ * Provides operations to add, remove, retrieve, and format tasks.
+ */
+public class TaskList {
+    private final ArrayList<Task> tasks;
```
robinlogic, line 11: Great job on following the coding Standard here!

src/main/java/capybara/command/ByeCommand.java
```diff
@@ -0,0 +1,17 @@
...
+import capybara.Ui;
+
+public class ByeCommand extends Command {
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        ui.showGoodbye();
+    }
+
+    @Override
+    public boolean isExit() {
```
robinlogic, line 36: Good job on following the coding standard!

src/test/java/capybara/TaskFormatTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class TaskFormatTest {
+
+    @Test
+    public void deadline_dateOnly_toFileString_hasDateOnly() {
+        LocalDateTime by = LocalDate.of(2025, 6, 4).atStartOfDay();
+        Task d = new Deadline("return book", by);
+        String file = d.toFileString();
+        // expected shape: D | 0 | return book | 2025-06-04
+        Assertions.assertTrue(file.matches("^D \\| [01] \\| return book \\| 2025-06-04$"),
```
robinlogic, line 17: Good job using regex to ensure testing flexibility!

src/test/java/capybara/command/MarkCommandTest.java
```diff
@@ -0,0 +1,50 @@
...
+        Command cmd = new MarkCommand(0, true);
+        cmd.execute(tasks, ui, storage);
+
+        Assertions.assertTrue(tasks.get(0).isDone(), "Task should be marked done");
+    }
+
+    @Test
+    public void execute_mark_outOfBounds_throwsCapyExceptionWithNiceMessage() throws IOException {
+        TaskList tasks = new TaskList(); // empty
+        Storage storage = new Storage(tmp.resolve("tasks.txt").toString());
```
robinlogic, line 39: If you are testing `TaskList`'s functionality, it might be worth to consider creating dummy classes for `Storage` and `Ui` to ensure this test is isolated from the issues in those classes.

src/main/java/capybara/CapyException.java
```diff
@@ -0,0 +1,15 @@
+package capybara;
+
+public class CapyException extends Exception {
```
x3zapper, line 9: Good job following the class naming conventions! 

Here is a food for thought, consider extending a more specific exception (e.g. `RuntimeException`) to better represent why this exception may be thrown. This may not matter as much in a small personal project but I do believe it will help when working in a large project where it may be infeasible to simply catch all of the custom exceptions defined in the project.

src/main/java/capybara/Capybara.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param filePath Path to the storage file where tasks are saved.
+     */
+    public Capybara(String filePath) {
+        this.ui = new Ui();
+        this.storage = new Storage(filePath);
+
+        TaskList loaded;
+        try {
+            loaded = new TaskList(storage.load());
+        } catch (Exception e) {
```
x3zapper, line 31: Similar to my earlier comment, it is generally good practice to be specific when using exceptions. For example, in this case this try/catch block may catch more than you initially designed for and is a typical trait of exception swallowing.

src/main/java/capybara/Capybara.java
```diff
@@ -0,0 +1,67 @@
...
+public class Capybara {
+    private final Storage storage;
+    private final TaskList tasks;
+    private final Ui ui;
+
+    /**
+     * Constructs a new Capybara instance with the given storage file path.
+     *
+     * @param filePath Path to the storage file where tasks are saved.
+     */
```
x3zapper, line 23: Good job on the detailed javadoc comment throughout!

src/main/java/capybara/Deadline.java
```diff
@@ -0,0 +1,23 @@
+package capybara;
+
+import java.time.LocalDateTime;
+
+public class Deadline extends Task {
+    private final LocalDateTime by;
+
+    public Deadline(String name, LocalDateTime by) {
+        super(name);
+        this.by = by;
```
x3zapper, line 24: Although using a short variable name is acceptable for short-lived variables (argument var name), you may want to consider using a more descriptive name for the instance level variable.

src/main/java/capybara/EmptyDescriptionException.java
```diff
@@ -0,0 +1,7 @@
+package capybara;
+
+public class EmptyDescriptionException extends CapyException {
+    public EmptyDescriptionException(String kind) {
+        super(String.format("Peep! %s needs a description. Try: %s sleep ...", kind, kind.toLowerCase()));
```
x3zapper, line 18: Good job on following the line length requirements!

src/main/java/capybara/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+import capybara.command.AddToDoCommand;
+import capybara.command.ByeCommand;
+import capybara.command.Command;
+import capybara.command.DeleteCommand;
+import capybara.command.FindCommand;
+import capybara.command.ListCommand;
+import capybara.command.MarkCommand;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
```
x3zapper, line 14: Well ordered imports, good job

src/main/java/capybara/Parser.java
x3zapper, line 1: Interesting that this parser is monolithic considering that you do have a `*Command` abstraction present. Would be good to consider moving more code into the command abstractions for a possibly cleaner and maintainable code.

src/main/java/capybara/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+                return LocalDateTime.parse(t, DATE_TIME);
+            }
+        } catch (Exception e) {
+            if ("/from".equals(label)) {
+                throw new CapyException("capybara.Capybara can’t read the /from time. Try 'yyyy-MM-dd' or 'yyyy-MM-dd HH:mm'.");
+            } else if ("/to".equals(label)) {
+                throw new CapyException("capybara.Capybara can’t read the /to time. Try 'yyyy-MM-dd' or 'yyyy-MM-dd HH:mm'.");
+            } else {
+                throw new CapyException("capybara.Capybara can’t read that date 🐹🍊. Try 'yyyy-MM-dd' or 'yyyy-MM-dd HH:mm'.");
+            }
```
x3zapper: Side note: If I am not mistaken, we are supposed to store `LocalDateTime` if applicable as it was never guaranteed that a user would conform to the date format.

src/main/java/capybara/Parser.java
```diff
@@ -0,0 +1,191 @@
...
+        return new DeleteCommand(idx0);
+    }
+
+    private static int parseOneBasedIndex(String raw, String kind) throws CapyException {
+        if (raw == null || raw.isBlank()) {
+            throw new EmptyTaskNumberException(kind);
+        }
+
+        try {
+            int n = Integer.parseInt(raw.trim());
```
x3zapper, line 202: A better option would be to name `n` as `index` as this generally makes things clearer to any reviewer and even the writer.

src/main/java/capybara/Storage.java
```diff
@@ -0,0 +1,140 @@
...
+     * the expected format will be skipped.
+     *
+     * @return an {@code ArrayList<capybara.Task>} containing the tasks read from file
+     * @throws IOException if an I/O error occurs while creating or reading the file
+     */
+    public ArrayList<Task> load() throws IOException {
+        java.time.format.DateTimeFormatter DATE_FMT =
+                java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;                // yyyy-MM-dd
+        java.time.format.DateTimeFormatter DATETIME_FMT =
+                java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"); // yyyy-MM-dd 18:00
```
x3zapper: Looking at the use of these variables, it may be a better option to declare them as a class static variable. E.g.: `private static final DateTimeFormatter VAR_NAME = ...`

x3zapper: Using the fully qualified package+class name `java.time.format.DateTimeFormatter` as the type may not be very clear to the average user.

It would be better to import at the top level and use only the class name here. This can also help with maintainability as you would know at a glance by looking at the code file header that you do use `DateTimeFormatter`.

src/main/java/capybara/Storage.java
x3zapper: This line is 135 chars long which is > than the recommended 110 (soft limit)/120 (max). Personally, I do try to keep my code at <=90 chars which helps with readability especially if one does not have a high resolution screen.

x3zapper (commented):
Overall good work on following the Java Coding Standard required by our course. I looked through about half of the PR in detail and left some comments/recommendations that you may wish to pursue.

PR #397, opened 2025-08-27
[Kineeeeee] iP

Kineeeeee:
# 🌸 Sora Chatbot Pull Request

This PR introduces the **Sora chatbot**, a simple task manager inspired by Duke.  
It can handle tasks such as *ToDo*, *Deadline*, and *Event*, and supports persistent storage. 😄

---

## ✨ Features Implemented
- Add `Todo`, `Deadline`, and `Event` tasks
- Mark and unmark tasks
- List all tasks
- Delete tasks
- Find tasks by keyword
- Save and load tasks from file
- Error handling with custom `SoraException`

---

## 🔢 Steps to Use
1. Clone the repository
2. Run `./gradlew build`
3. Start the app with `java -jar build/libs/sora.jar`
4. Type `list` to view tasks
5. Type `bye` to exit

---

## 💻 Example Code

```java
    Sora sora = new Sora("data/tasks.txt");
    sora.run();
}
```

---

## Progress
- [x] Implement `Task`
- [x] Add unit tests
- [x] Update documentation


src/main/java/sora/list/TaskList.java
```diff
@@ -0,0 +1,110 @@
...
+    public int size() {
+        return tasks.size();
+    }
+
+    /**
+     * Returns the full list of tasks.
+     *
+     * @return the list of tasks as an {@link ArrayList}.
+     */
+    public ArrayList<Task> getFullTasks() {
```
ravichandran-gokool, line 72: Maybe it would be clearer if the name of this method is getAllTasks() as the one could possibly misunderstand this method to be a getter for 'full tasks' 🙇

src/main/java/sora/task/Deadline.java
```diff
@@ -0,0 +1,43 @@
+package sora.task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+/**
+ * Represents a {@code Deadline} task that must be completed by a specific date and time.
+ */
+public class Deadline extends Task {
```
ravichandran-gokool, line 10: For the different types of tasks that extend Task, perhaps it would be good to name the class XxxxxTask, so in this case DeadlineTask might be a better name 👍

ravichandran-gokool (commented):
Overall very good names given to fields and methods! Good job!

build.gradle
Cavannn, line 1: Overall, the code follows Java conventions well — naming is mostly consistent and the project is easy to navigate. With a few refinements (especially around method size and documentation), the code will align even more closely with Java best practices.

src/main/java/sora/task/Task.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Represents a task in the Sora task management system.
+ * This is the abstract base class for specific types of tasks:
+ * {@code Todo}, {@code Deadline}, and {@code Event}.
+ */
+public class Task {
+    /**
+     * Enum representing the type of the task.
+     */
+    public enum TaskType {
```
Cavannn, line 12: The TaskType enum is currently nested inside the Task class. This is fine, but if it’s referenced frequently outside of Task, you might consider moving it to its own file for better modularity.

src/main/java/sora/task/Task.java
```diff
@@ -0,0 +1,110 @@
...
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
+
+    /**
+     * Returns a user-friendly string representation of the task.
+     *
+     * @return the formatted string with task type, status, and description.
+     */
+    public String toString() {
```
Cavannn, line 80: Since you’re overriding toString(), you should explicitly annotate it with @Override. This helps catch mistakes if the method signature changes in the future.

PR #416, opened 2025-08-28
Itsdaodao iP

Itsdaodao:
# George 🐒

> "Why go bananas trying to remember your tasks when George can do it for you?" - Some monkey, probably George

George is your personal monkey assistant that helps to take over the boring stuff.

It's:
- **text-based**
-*easy to learn*
- ~~slow~~ **SUPER FAST** to use

All you need to do is:
1. download it from [here](https://github.com/Itsdaodao/ip/releases/tag/A-Jar)
2. double-click it
3. add your bananas (tasks) 🍌
4. let George handle the rest

And the best part? It's *FREE!**

### Features
- [x] Managing tasks  
- [ ] Reminders (coming soon)  

---

If you’re a Java programmer, you can use it to practice too! Here’s the `main` method:  

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/TaskManager.java [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+public class TaskManager {
+    private List<Task> tasksList;
+    private Storage storage;
+
+    public TaskManager(String dataFileName) {
+        this.tasksList = new ArrayList<>();
+        this.storage = new Storage(dataFileName);
+    }
+
+    public void addToDo(String description) throws GeorgeException {
```
bearkerb: Would it be better to rename this to addToDoTask to remain consistent with the next two methods with similar purposes?

src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public DeadlineTask(String description, String deadline, boolean isDone) throws GeorgeException {
+        super(description);
+        this.deadline = deadline;
+        this.isDone = isDone;
+    }
+
+    @Override
+    public String getType() {
```
bearkerb: Would it be better to change the method name to something like getTypeAsString() as I feel it's not too intuitive that this function returns a String. (same for other task classes)

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,86 @@
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TaskManager {
+    private List<Task> tasksList;
```
bearkerb: I like that you make it clear with the naming that this is a multi-valued variable

src/main/java/Task.java
```diff
@@ -0,0 +1,35 @@
+abstract class Task {
+    protected String description;
+    protected boolean isDone;
```
bearkerb: I like that you make it clear you have a boolean variable here with the naming.

bearkerb (commented):
Nice work overall. 😎

src/main/java/George.java
```diff
@@ -0,0 +1,85 @@
...
+        System.out.println(newLine);
+    }
+
+    private static TaskManager start() throws IOException {
+        TaskManager manager = new TaskManager(fileName);
+        manager.load();
+        return manager;
+    }
+
+    private static void echo(TaskManager manager) {
```
Retribob, line 39: While it is a private method, it would be more intuitive if the method name was changed to better reflect what it does. Maybe something like inputParser or smt.

PR #419, opened 2025-08-28
[mellowfarm] iP

mellowfarm:
### **Bug**

> “A bug in the code is just a feature waiting to be fixed!” 🐛💻

Bug frees your mind from having to remember tasks. It’s simple, text-based, and designed to manage your to-dos.

**Pros**

1. Text-based
2. Easy-to-learn
3. SUPER fast to use!

**All you need to do is...**

1. Download it from here.
2. Run the application.
3. Add your tasks.
4. Let Bug manage your tasks for you!

And the best part is... It's free!! 🎉

**Features**

- [ ] Managing tasks 
- [ ] Reminders (coming soon)

`public class Main {
        Application.launch(Bug.class, args);
    }
}`


build.gradle
```diff
@@ -0,0 +1,50 @@
...
+        )
+    }
+}
+
+shadowJar {
+    archiveBaseName = "bug"
+    archiveClassifier = null
+}
+
+run{
```
nghnaomi: Maybe add a space between "run" and the opening bracket?

src/main/java/bug/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+            this.end = null;
+            this.index = -1;
+            this.error = ":(! empty input!";
+            return;
+        } else {
+            String[] split = input.split("\\s+", 2);
+            String instruction = split[0];
+            String contents = (split.length > 1) ? split[1] : "";
+
+            switch(instruction) {
```
nghnaomi: Add space between "switch" and opening bracket?

src/main/java/bug/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+                case "find":
```
nghnaomi: Should this be `case FIND:` and similarly for the rest?

nghnaomi (commented):
LGTM. Just a few nits to fix.

src/main/java/bug/Bug.java
```diff
@@ -0,0 +1,141 @@
+package bug;
+
+/**
+ * The Bug class is the entry point of the task management application.
+ * It interacts with the user, manages tasks, and stores them in a file.
+ */
+public class Bug {
```
Habutiii, line 18: What a nice name for your bot 🤣

src/main/java/bug/Bug.java
```diff
@@ -0,0 +1,141 @@
...
+                    ui.showBye();
+                    return;
+                }
+                case UNKNOWN: {
+                    ui.showError(parser.error);
+                    break;
+                }
+                default:
+                    ui.showError(":(! i don't know what you mean! please re-enter your task :)!");
+            }
```
Habutiii: For non-lambda style swith case should not have indentation for your case as per the coding standard

This is the enforced coding standard for switch-case:
```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

You may find it [here](https://se-education.org/guides/conventions/java/intermediate.html)

src/main/java/bug/Bug.java
```diff
@@ -0,0 +1,141 @@
...
+            if (task.getDescription().toLowerCase().contains(keyword.toLowerCase())) {
+                ui.showTasks(i + 1, task);
+                found = true;
+            }
+        }
+        if (!found) {
+            ui.showError(":(! no tasks match the keyword: " + keyword + "!");
+        }
+        ui.showLine();
+    }
```
Habutiii: Since you have a handleFind as a function that makes it inconsistent with other commands, would it be better if we extract all commands to a single Command.java with public functions that handle each command?

eg. Command.handleFind(), Command.handleList(), etc.

If you would like to do the `More-OOP` task, can split each command to individual classes.

src/main/java/bug/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+        if (input == null || input.isBlank()) {
+            this.type = Type.UNKNOWN;
+            this.desc = null;
+            this.by = null;
+            this.start = null;
+            this.end = null;
+            this.index = -1;
+            this.error = ":(! empty input!";
+            return;
+        } else {
```
Habutiii: Maybe throwing an exception here can save you many lines of code here?

src/main/java/bug/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+public class Parser {
+    private static final DateTimeFormatter INPUT_DT = DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+    private static final DateTimeFormatter INPUT_DT2 = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
+    public enum Type { LIST, TODO, DEADLINE, EVENT, MARK, UNMARK, DELETE, BYE, UNKNOWN, FIND };
+
+    public Type type;
+    public String desc;
+    public LocalDate by;
+    public LocalDateTime start, end;
```
Habutiii: Having command specific parameters resolved in the parser would make it very bulky when you have more commands and parameters in the future. 

Would it be better to leave the handling of command specific parameters in the command logic itself?

A suggestion would be just identify the command type in the parser then leave the parameters as a single string, resolve and handle it while you are handling the command logic.

src/main/java/bug/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+                        LocalDateTime start = LocalDateTime.parse(p[3]);
+                        LocalDateTime end = LocalDateTime.parse(p[4]);
+                        Task event = new Events(p[2], start, end);
+                        if (isDone) {
+                            event.markAsDone();
+                        }
+                        out.add(event);
+                        break;
+                    default:
+                        break;
```
Habutiii: For non-lambda style swith case should not have indentation for your case as per the coding standard

This is the enforced coding standard for switch-case:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```
You may find it [here](https://se-education.org/guides/conventions/java/intermediate.html)

src/main/java/bug/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+                case "unmark":
+                case "delete": {
+                    this.type = instruction.equals("mark") ? Type.MARK : instruction.equals("unmark")
+                            ? Type.UNMARK : Type.DELETE;
+                    this.index = Integer.parseInt(contents) - 1;
+                    break;
+                }
+                default:
+                    this.type = Type.UNKNOWN;
+                    this.error = ":(! i don't know what you mean! please re-enter your task :)!";
```
Habutiii: For non-lambda style swith case should not have indentation for your case as per the coding standard

This is the enforced coding standard for switch-case:

```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```
You may find it [here](https://se-education.org/guides/conventions/java/intermediate.html)

Habutiii (commented):
Your project really has an interesting name.

Overall coding style and naming conventions are ok. Just be more careful for the switch-case indentations.

Also, it might be better to let your parser handle errors using the custom exception instead of the error attribute.

Well Done!

PR #423, opened 2025-08-28
[Eldridge Ng] iP

eltry42:
# 🚀 Introducing Eltry – Your Friendly Task Manager Bot

This PR adds **Eltry**, a lightweight, text-based task manager that helps you capture and organize your to-dos, deadlines, and events — so you can free your mind and stay focused.

> “Your mind is for having ideas, not holding them.” – David Allen

## ✨ Features
- ✅ **Text-based & fast** – no distractions, just productivity
- ✅ **Easy to learn, quick to use**
- ✅ **Saves tasks to file** – nothing is lost
- 🚧 **Deadlines & events** – fully supported
- 🚧 **Reminders** – coming soon!
- 💡 **Great for Java learners** – built with OOP, Gradle, and JUnit

## 📦 How to Use
1. Download  from [here](#) 
2. Add tasks like:
   - `todo read book`
   - `deadline submit report /by 2025-04-05 1800`
   - `event team meeting /from 0900 /to 1000`
3. Use `list`, `mark`, `delete`, and `bye`

## 💻 For Java Developers
Want to learn or contribute? Check out the main class:

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/eltry/Command.java
Comment (JovanTeo1103) [inline src/main/java/eltry/Command.java:1]: Good use of constructor overloading to support different kinds of commands.

You might want to consider whether having arg1,2,3 as separate fields is flexible enough. what if a command requires 4+ commands in the future. Maybe using a List<String> would scale better.

File: src/main/java/eltry/Deadline.java
Comment (JovanTeo1103) [inline src/main/java/eltry/Deadline.java:1]: Good exception handling for invalid date input, giving a helpful error message.

In toString(), toLowerCase() is applied to the entire formatted date string, this might make “AM/PM” less readable. You could consider only lowercasing the month or removing it.

File: src/main/java/eltry/Eltry.java
Comment (JovanTeo1103) [inline src/main/java/eltry/Eltry.java:1]: Good separation of concerns, your UI storage and parsers are used correctly.

The main method however is quite long, consider refactoring parts of it into helper methods to improve readability.

File: src/main/java/eltry/Event.java
Comment (JovanTeo1103) [inline src/main/java/eltry/Event.java:1]: Exception handling for invalid date input is user-friendly.

Consider overriding equals() and hashCode() if Event objects might be compared or stored in collections.

File: src/main/java/eltry/Command.java [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+public class Command {
+
+    /** The primary action/command keyword (e.g., "todo", "deadline", "list"). */
+    public final String action;
+
+    /** The index associated with commands that operate on a task (e.g., mark/unmark/delete). */
+    public final int index;
+
+    /** First argument for commands that require additional input. */
+    public final String arg1;
```
soomin-k, line 18: i think the naming could be improved to be more specific in what it does instead of arg1, 2, and 3!😇

src/main/java/eltry/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Returns a human-readable string representation of the Deadline task.
+     *
+     * @return formatted string including task type, description, and due date/time
+     */
+    @Override
+    public String toString() {
+        String pretty = by.format(DateTimeFormatter.ofPattern("MMM dd yyyy, h:mma")).toLowerCase();
```
soomin-k, line 44: i think the naming can be improved, as it may be misleading ! maybe formattedStr or smt like that ?

src/main/java/eltry/Eltry.java
```diff
@@ -0,0 +1,114 @@
...
+                        storage.save(taskList.getAll());
+                        ui.showTaskMarked(taskList.get(cmd.index));
+                        break;
+                    case "unmark":
+                        taskList.unmark(cmd.index);
+                        storage.save(taskList.getAll());
+                        ui.showTaskUnmarked(taskList.get(cmd.index));
+                        break;
+                    case "delete":
+                        Task removed = taskList.remove(cmd.index);
```
soomin-k: maybe you could name it removedTask!

src/main/java/eltry/Parsers.java
```diff
@@ -0,0 +1,86 @@
...
+public class Parsers {
+
+    /**
+     * Parses a user input string into a {@link Command}.
+     *
+     * @param input the raw user input string
+     * @return the parsed Command object
+     * @throws EltryException if the input is invalid or the format is incorrect
+     */
+    public Command parse(String input) throws EltryException {
```
soomin-k: i think in general the name of the variables can be clearer and longer since now it is very simplified ! :)

PR #433, opened 2025-08-28
[JovanTeo1103] iP

JovanTeo1103:
# BaymaxPro  

> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://gettingthingsdone.com/))  

BaymaxPro frees your mind of having to remember things you need to do. It's,  

- text-based  
- easy to learn  
- **FAST SUPER FAST** to use  
- ~~boring~~ fun to use 😉  

All you need to do is,  

1. download it from [here](https://github.com/JovanTeo1103/ip/releases/tag/A-Jar)  
2. double-click it  
3. add your tasks  
4. let it manage your tasks for you 🚀  
5. enjoy 🎉  

And it is *FREE*!  

---

## Features  

- [x] Managing tasks  
- [x] Managing deadlines (coming soon)  
- [ ] Reminders (coming soon)  

```java
public class Main extends Application {

    private Baymax baymax = new Baymax();

    @Override
    public void start(Stage stage) {
        try {
            stage.setMinHeight(220);
            stage.setMinWidth(417);

            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
            AnchorPane ap = fxmlLoader.load();
            Scene scene = new Scene(ap);
            stage.setScene(scene);
            fxmlLoader.<MainWindow>getController().setBaymax(baymax);  // inject the Duke instance
            stage.show();
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}

### CODE REVIEW

File: src/main/java/baymax/Baymax.java [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+
+    public void run() {
+        Scanner scanner = new Scanner(System.in);
+        ui.showWelcome();
+
+        while (true) {
+            String input = scanner.nextLine();
+            try {
+                String command = Parser.getCommand(input);
+                String args = Parser.getArgs(input);
```
jireh0108: Would be better if Parser returns Command objects instead of raw strings -> makes it much easier to extend the code when adding new commands and cleaner code overall.

src/main/java/baymax/Baymax.java
```diff
@@ -0,0 +1,95 @@
...
+                    ui.showTaskAdded(t, tasks.size());
+                    storage.save(tasks.getAll());
+                }
+                case "find" -> {
+                    String keyword = input.substring(5).trim();
+                    ArrayList<Task> matches = tasks.findTasks(keyword);
+                    ui.showFoundTasks(matches);
+                }
+                default -> ui.showError("OHNO!!! I'm sorry, but that's not a valid command T.T");
+                }
```
jireh0108: Boilerplate code for every case is repetitive, would be good to have a helper function or better yet have it under a Command.execute(Ui, Storage, TaskList) method.

Also would be better to keep the switch logic in parser instead of leaving it in the main function (which should be meant only for running the application).

src/main/java/baymax/Event.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class Event extends Task {
+
+    private final LocalDate from;
+    private final LocalDate to;
+
+    public Event(String taskName, TaskType taskType, String fromStr, String toStr) {
+        super(taskName, taskType);
+        this.from = LocalDate.parse(fromStr);
+        this.to = LocalDate.parse(toStr);
```
jireh0108: This is dangerous because if the date String is not valid, a DateTimeParseException will be thrown which is not explicitly handed. This risks causing exceptions to bubble up and crash the program. Instead this should be placed in a try-catch block to catch any runtime exceptions.

src/main/java/baymax/Deadline.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Constructs a Deadline task with the given name, type, and due date.
+     *
+     * @param taskName the description of the task
+     * @param taskType the type of the task (should be TaskType.DEADLINE)
+     * @param byStr    the due date as a string in ISO_LOCAL_DATE format (yyyy-MM-dd)
+     */
+    public Deadline(String taskName, TaskType taskType, String byStr) {
+        super(taskName, taskType);
+        this.by = LocalDate.parse(byStr);
```
jireh0108: See comment for baymax.Event

src/main/java/baymax/InvalidCommandException.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public class InvalidCommandException extends Exception {
+
+    /**
+     * Constructs a new InvalidCommandException with the specified detail message.
+     *
+     * @param message the detail message explaining the reason for the exception
+     */
+    public InvalidCommandException(String message) {
+        super(message);
```
jireh0108: There's no difference in behaviour between InvalidCommandException and InvalidDescriptionException, the error message for each type of exception should be placed inside the classes. Additionally, it would be better if both exception classes inherited from a custom Exception class (e.g. BaymaxException) so that you can customise your exceptions easily.

src/main/java/baymax/Parser.java
```diff
@@ -0,0 +1,65 @@
+package baymax;
+
+public class Parser {
+
+    /** Extracts the command keyword from the raw input */
+    public static String getCommand(String input) {
+        String[] parts = input.trim().split(" ", 2);
+        return parts[0];
+    }
```
jireh0108: It would be better if getCommand() returns a Command class, with each type of command being a Command subclass - your code will be cleaner and extending it with a new command would be much easier to do.

src/main/java/baymax/TaskType.java
```diff
@@ -0,0 +1,7 @@
+package baymax;
+
+public enum TaskType {
+    TODO,
+    DEADLINE,
+    EVENT
+}
```
jireh0108: Similar to other comments, should make individual classes for each TaskType, all extending a Command class.

jireh0108 (commented):
Overall good job, just some tweaks about class responsibilities and error handling needed.

src/main/java/baymax/Command.java
```diff
@@ -0,0 +1,8 @@
+package baymax;
```
adibice1: might be better to leave the java classes under OOP folders (eg Command, Task, UI, Parser)

src/main/java/baymax/Deadline.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a deadline-type task in the Baymax task management system.
+ * A Deadline task has a description, a status (done or not done), and a
+ * due date represented as a {@link LocalDate}.
+ * <p>
+ * This class extends {@link Task} and provides functionality to
```
adibice1: I like the headers

src/main/java/baymax/Deadline.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param taskName the description of the task
+     * @param taskType the type of the task (should be TaskType.DEADLINE)
+     * @param byStr    the due date as a string in ISO_LOCAL_DATE format (yyyy-MM-dd)
+     * @param isDone   the completion status of the task
+     */
+    public Deadline(String taskName, TaskType taskType, String byStr, boolean isDone) {
+        super(taskName, taskType, isDone);
+        this.by = LocalDate.parse(byStr);
+    }
+
```
adibice1: should put in try catch blocks to introduce exceptions (for eg when string is empty etc)

src/main/java/baymax/FindCommand.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class FindCommand extends Command {
+    private final String keyword;
+
+    public FindCommand(String keyword) {
+        this.keyword = keyword;
+    }
+
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
```
adibice1: can introduce exceptions as well like the other classes

adibice1 (commented):
look good :) just need to introduce exceptions to catch bugs

PR #437, opened 2025-08-28
[rizrn] iP

rizrn:
# Woop bot
The most _adorable_ task tracker to date 😊 
> Upah! Feed me some tasks!

It's
- fast
- efficient
- cute

Want to try? [Download](https://github.com/rizrn/ip/releases/tag/A-Jar) **now**!

## Woop's abilities
Track 3 types of tasks for you
1. to do
2. deadline
3. event

## Easy to build new features
You can add new commands! Just add them to this switch statement in `WoopLogic`
```java
switch (userInput) {
case "delete":
    // logic
}
```

- [X] Manage Tasks
- [ ] GUI 
- [ ] Deadline notifications


src/main/java/WoopAI/DeadlineTask.java
```diff
@@ -0,0 +1,79 @@
+package WoopAI;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+
+/**
+ * A class for tasks of deadline type.
+ */
+public class DeadlineTask extends Task{
```
CZX123:
```suggestion
public class DeadlineTask extends Task {
```
Coding standard violation

src/main/java/WoopAI/IllegalDescriptorException.java
```diff
@@ -0,0 +1,7 @@
+package WoopAI;
+
+/**
+ * Exception for when an input command does not fit the correct description needed to run.
+ */
+public class IllegalDescriptorException extends Exception{
```
CZX123, line 6:
```suggestion
public class IllegalDescriptorException extends Exception {
```

data/data.txt
CZX123, line 1: This file should be deleted, can be done with the GitHub web UI directly so .gitignore doesn't have to be changed again.

src/main/java/WoopAI/DeadlineTask.java
```diff
@@ -0,0 +1,79 @@
...
+     * dueDate is parsed into a LocalDate.
+     *
+     * @param name The name of task.
+     * @param isFinished Whether task is finished.
+     * @param dueDate The due date of task.
+     */
+    public DeadlineTask(String name, boolean isFinished, String dueDate) {
+        super(name, isFinished);
+        this.dueDate = LocalDate.parse(dueDate);
+    }
```
CZX123: Strange that this constructor does not throw an exception when the previous one does.

src/main/java/WoopAI/DeadlineTask.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private String getDateString() {
+        return this.dueDate.getMonth().toString().substring(0, 3)
+                + " " + this.dueDate.getDayOfMonth()
+                + " " + this.dueDate.getYear();
```
CZX123, line 50: There is a more concise way to represent date strings, look into the date `format` and `DateTimeFormatter.ofPattern` methods.

src/main/java/WoopAI/EventTask.java
```diff
@@ -0,0 +1,69 @@
+package WoopAI;
+
+/**
+ * A class for event type task.
+ */
+public class EventTask extends Task{
```
CZX123:
```suggestion
public class EventTask extends Task {
```

src/main/java/WoopAI/DeadlineTask.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
CZX123, line 55: Unnecessary javadoc.

src/test/java/WoopAI/ParserTest.java
```diff
@@ -0,0 +1,30 @@
+package WoopAI;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
CZX123: Same as above.

text-ui-test/input.txt
CZX123, line 1: Should add find command here as well. It's good practice to add any new features to the test immediately so any regression is caught much earlier.

src/main/java/WoopAI/WoopLogic.java
```diff
@@ -0,0 +1,85 @@
...
+                    break;
+                default:
+                    throw new UnknownCommandException();
+                }
+            } catch (UnknownCommandException e) {
+                Ui.showCommandError(command);
+            } catch (IllegalDescriptorException e) {
+                Ui.showDescriptorError(command);
+            } catch (Exception e) {
+                Ui.showError(e);
```
CZX123: Pokemon exception handling is heavily discouraged, should define a custom exception class for the project and use that instead, so actual coding errors are not hidden.

src/main/java/WoopAI/WoopLogic.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Scanner;
+
+/**
+ * Handles main logic of Woop
+ */
+public class WoopLogic {
+    private final Scanner sc;
+    private static final String EXIT = "bye"; //String constant to exit loop
+    private TaskList taskList;
```
CZX123: Should list the static fields before the member fields, it's part of the coding standard as an advanced rule.

src/main/java/WoopAI/TodoTask.java
```diff
@@ -0,0 +1,53 @@
+package WoopAI;
+
+/**
+ * A class for to do type tasks
+ */
+public class TodoTask extends Task{
```
CZX123:
```suggestion
public class TodoTask extends Task {
```

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Retrieves the index of the task from the user input.
+     * The index of the task is used for commands that require to edit at specific indexes.
+     *
+     * @param input The user input.
+     * @return The index of the Task in int.
+     * @throws IllegalDescriptorException If descriptor syntax is not valid.
+     */
+    public static int parseIndex(String input) throws IllegalDescriptorException {
+        if (!input.contains(" ")) {
```
CZX123: The condition here is not exhaustive, could use try catch with Integer.parseInt, and also remember to account for indices that are out-of-bounds.

src/main/java/WoopAI/EventTask.java
```diff
@@ -0,0 +1,69 @@
...
+    public EventTask(String name, boolean isFinished,
+                     String startTime, String endTime) {
+        super(name, isFinished);
+        this.startTime = startTime;
+        this.endTime = endTime;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
CZX123, line 44:
```suggestion
```

text-ui-test/runtest.bat
CZX123, line 1: Should add a command here that removes the data file so the tests are consistent.

src/test/java/WoopAI/EventTaskTest.java
```diff
@@ -0,0 +1,25 @@
+package WoopAI;
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
```
CZX123, line 7: Same as above.

src/test/java/WoopAI/DeadlineTaskTest.java
```diff
@@ -0,0 +1,45 @@
+package WoopAI;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
CZX123, line 9: Import order violates coding standard, static imports should go before the rest.

src/main/java/WoopAI/TodoTask.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param name The name of task.
+     * @param isFinished Whether task is finished.
+     */
+    public TodoTask(String name, boolean isFinished) {
+        super(name, isFinished);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
CZX123, line 33:
```suggestion
```

src/main/java/WoopAI/Task.java
```diff
@@ -0,0 +1,97 @@
...
+     * @return Encoded string for save state.
+     */
+    public String getSaveInfo() {
+        return " | " + (this.isFinished ? "1" : "0")
+                + " | " + this.name;
+    };
+
+    public boolean containsText(String text) {
+        return this.name.contains(text);
+    }
```
CZX123, line 75: Missing javadoc.

.gitignore
CZX123, line 1: Good job identifying that the data file shouldn't be committed.

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param input The user input.
+     * @param type The Task type that this descriptor is for.
+     * @return A new Task object that represents the descriptor.
+     * @throws IllegalDescriptorException If descriptor syntax is invalid.
+     * @throws UnknownCommandException If type given is invalid.
+     */
+    public static Task parseDescriptor(String input, TaskType type)
+            throws IllegalDescriptorException, UnknownCommandException {
+        int index = input.indexOf(" ");
+        if (index >= 0 && (index + 1) < input.length()) {
```
CZX123:
```suggestion
        if (0 <= index && index < input.length() - 1) {
```
This is subjective, but I personally find this to be more readable and clearer, feel free to discuss.

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+            String descriptor =  input.substring(index + 1);
+            switch (type) {
+            case TODO:
+                return new TodoTask(descriptor);
+            case EVENT:
+                String[] tmp = descriptor.split(" /from ");
+                String eventSubject = tmp[0];
+                String[] tmp2 = tmp[1].split(" /to ");
+                String startTime = tmp2[0];
+                String endTime = tmp2[1];
```
CZX123: Names here can be more descriptive despite being temporary. I suggest `subjectIntervalSplit` instead of `tmp` and `startEndSplit` instead of `tmp2`.

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+                return new EventTask(eventSubject, startTime, endTime);
+            case DEADLINE:
+                tmp = descriptor.split(" /by ");
```
CZX123: Same as above. Suggest `subjectDateSplit` instead of `tmp`. In this case, it also prevents the issue of re-using the `tmp` variable which might introduce bugs later on.

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param input The user input.
+     * @return The index of the Task in int.
+     * @throws IllegalDescriptorException If descriptor syntax is not valid.
+     */
+    public static int parseIndex(String input) throws IllegalDescriptorException {
+        if (!input.contains(" ")) {
+            throw new IllegalDescriptorException();
+        }
+        String tmp = input.split(" ")[1];
```
CZX123: Can just name it `index` instead of `tmp`.

src/main/java/WoopAI/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+                boolean isFinished = tmp[1].equals("1");
+                switch (tmp[0]) {
+                case "D":
+                    String dueDate = tmp[3];
+                    tasks.addTask(new DeadlineTask(
+                            name, isFinished, dueDate));
+                    break;
+                case "E":
+                    String startTime = tmp[3];
+                    String endTime = tmp[4];
```
CZX123: Naming as well. Can just be `lineSplit` instead of `tmp`.

src/main/java/WoopAI/Task.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param isFinished True if task is finished, false otherwise.
+     */
+    public Task(String name, boolean isFinished) {
+        this.name = name;
+        this.isFinished = isFinished;
+    }
+
+    public boolean getIsFinished() {
+        return this.isFinished;
+    }
```
CZX123, line 40: I believe it's fine to leave the method name as `isFinished`, same as the field name. Calling `getIsFinished` may feel awkward.

src/main/java/WoopAI/Task.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public String toString() {
+        String finished = "[X]";
+        String unfinished = "[ ]";
+
+        if (this.isFinished) {
+            return finished + " " + this.name;
+        } else {
+            return unfinished + " " + this.name;
+        }
```
CZX123, line 104: This can be more concise.

src/main/java/WoopAI/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+        } catch (IOException e) {
+            Ui.showIoError();
+        }
+    }
+
+    /**
+     * Saves tasks from the given TaskList object to data file
+     * If I/O error occurs, prints an error message.
+     *
+     * @param list Current TaskList to store.
```
CZX123, line 38:
```suggestion
     * @param list current TaskList to store.
```

src/main/java/WoopAI/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+        for (int i = 0; i < getSize(); i++) {
+            sb.append(this.tasks.get(i).getSaveInfo());
+            sb.append("\n");
+        }
+
+        sb.deleteCharAt(sb.length() - 1);
+        return sb.toString();
+    }
+
+    public TaskList findKeyword(String keyword) {
```
CZX123, line 105: Missing javadoc.

src/main/java/WoopAI/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+        TaskList resList = new TaskList();
+        for (Task task : this.tasks) {
+            if (task.containsText(keyword)) {
+                resList.addTask(task);
+            }
+        }
+        return  resList;
```
CZX123:
```suggestion
        return resList;
```

src/main/java/WoopAI/Ui.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public static void showTaskList(TaskList list) {
+        if (list.isEmpty()) {
+            System.out.println("Upah! Feed me some tasks to track!");
+        } else {
+            System.out.println("Upah! I have detected several tasks");
+            System.out.println(list);
+        }
+
```
CZX123:
```suggestion
```

src/test/java/WoopAI/ParserTest.java
```diff
@@ -0,0 +1,30 @@
...
+        Task t = Parser.parseDescriptor("todo hello fs", TaskType.TODO);
+        assertEquals(new TodoTask("hello fs"), t);
+    }
+
+    @Test
+    public void parseDescriptor_eventTask_taskReturned()
+        throws IllegalDescriptorException, UnknownCommandException {
+        Task t = Parser.parseDescriptor("event hello /from 2 pm /to 4 pm", TaskType.EVENT);
+        assertEquals(new EventTask("hello", "2 pm", "4 pm"),
+                t);
```
CZX123, line 28: Unnecessary line-break.

CZX123 (commented):
👍Overall well-written code that follows the coding standards.
However, some pointers:
- There are quite a lot of use of `this.field` in methods, which actually violates the advanced level of the coding standards.
- There are a lot of line breaks when the original line length is not exceeded yet, which if done tastefully can help improve readability. But there are many examples where I feel it makes it less readable, especially ones where the method name is longer than the 8-space indent, like:
  ```java

    public static void showDescriptorError(String command) {
        System.out.println("Upah! The command " + command
                + " needs a valid description!");
    }
   ```

src/main/java/WoopAI/TaskList.java
```diff
@@ -0,0 +1,120 @@
+package WoopAI;
+
+import java.lang.StringBuffer;
+import java.util.ArrayList;
+
+/**
+ * Class to keep track of all Task objects in current Woop run.
+ */
+public class TaskList {
+    private ArrayList<Task> tasks;
```
luciusyeo, line 9: Good job naming multi-valued noun!

src/main/java/WoopAI/TaskList.java
luciusyeo, line 105: Perhaps the parameter name could more clearly explain the purpose, since it may not be apparent it is used for searching tasks that contain specific words! Maybe "searchTerm" or "query"

luciusyeo, line 105: Also, the method name sounds like it is finding Keywords although it is actually returning tasks. Perhaps "findTasksWithKeyword" or "filterByKeyword" would be clearer

src/main/java/WoopAI/TaskList.java
luciusyeo: resList sounds slightly unclear to me. Perhaps rename this variable to something like "matchingTasks"

src/main/java/WoopAI/Ui.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public static void showFindKeyword(TaskList list) {
```
luciusyeo: Same issue here!

luciusyeo (commented):
LGTM. Just small nits!

src/main/java/WoopAI/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+                String dueDate = tmp[1];
+                return new DeadlineTask(deadlineSubject, dueDate);
+            default:
+                throw new UnknownCommandException();
+            }
+        } else {
+            throw new IllegalDescriptorException();
+        }
+    }
+
```
eesyuen09: Would names like subjectDateSplit / startEndSplit (or lineSplit) instead of tmp/tmp2 make the intent clearer? I noticed similar naming in a few places too.

eesyuen09 (commented):
I like your code structure overall. I noticed that several peers have already provided detailed feedback on coding standard and code quality issues. To avoid duplicating their comments, I won’t repeat the same points here.
Instead, I’d just like to highlight that I agree with the observations made about improving naming clarity (tmp, resList, findKeyword) and exception handling. These changes would definitely make the code easier to read and maintain.

PR #438, opened 2025-08-28
[Lucas Low] IP

BiggillyBoyMan:
NamiBot
-------------------------------------------------------------------------------------------------------------------------

NamiBot is a revolutionary todo list bot that helps you track deadlines, events and todo list. With features such as being able to mark your tasks, adding and removal of tasks. Nami bot helps users keep track of their tasks, ensuring users can keep track of their tasks.

### Why Nami?
- Text-based CLI + JavaFX GUI
- Simple commands (`list`, `todo`, `deadline`, `event`, `find`, `delete`)
- Saves to disk automatically

### Why Nami?
- Text-based CLI + JavaFX GUI
- Simple commands (`list`, `todo`, `deadline`, `event`, `find`, `delete`, `mark`, `unmark`)
- Auto-saves to disk (`./data/duke.txt`)

### Getting Started
1. Download the JAR from the **[latest release](https://github.com/BiggillyBoyMan/ip/releases)**.
2. Put it in an empty folder.
3. Open a terminal in that folder.
4. Run `java -jar "Nami.jar"`.
5. Add tasks and let Nami manage them for you!

### Example Session
```bash
# launch, add a todo, list tasks, then exit
java -jar "Nami.jar"
todo read book
list
bye

### CODE REVIEW

File: src/main/java/Nami/Nami.java [hunk 1]
```diff
@@ -0,0 +1,185 @@
+package Nami;
```
sahishnuk2: Package name should be in all lowercase

src/main/java/Nami/parser/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final DateTimeFormatter INPUT_DATE_FORMAT = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+
+    public static Command parse(String full) throws DukeException {
+        String input = full.trim();
+        String[] parts = input.split(" ", 2);
+        String keyword = parts[0].toLowerCase();
+        String args = parts.length > 1 ? parts[1].trim() : "";
+
+        switch (keyword) {
+            case "bye":
```
sahishnuk2, line 27: There should be no indentation for "case". Maybe, you can edit your IDE configurations to following the coding style.

src/main/java/Nami/task/Deadlines.java
```diff
@@ -0,0 +1,35 @@
+package Nami.task;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadlines extends Tasks {
```
sahishnuk2, line 5: I think it would be better of the class name is singular, as it is only a recipe for multiples object bring created. I noticed the same issues in the several other classes too

src/main/java/Nami/task/toDo.java
```diff
@@ -0,0 +1,22 @@
+package Nami.task;
+public class toDo extends Tasks {
```
sahishnuk2: Your class name should start with a uppercase letter

src/test/java/Nami/parser/ParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+        assertTrue(ex.getMessage().toLowerCase().contains("deadline"),
+                "Message should mention deadline problem");
+    }
+
+    @Test
+    void parse_deadline_valid_returnsAddDeadlineCommand() throws Exception {
+        Command c = Parser.parse("deadline return book /by 2/12/2019 1800");
+        assertTrue(c instanceof AddDeadlineCommand,
+                "Expected AddDeadlineCommand for correctly formatted deadline");
+    }
```
sahishnuk2, line 42: I like the test names as they are very descriptive 👍. Perhaps, a clearer way is to have 3 parts to each test name separated by underscores. (e.g. parse_deadlineValid_returnsAddDeadlineCommand) This comment is for all your test cases in your programme.

src/main/java/Nami/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,32 @@
+package Nami.command;
+import java.time.LocalDateTime;
```
sahishnuk2: Maybe you can place this import statement on line below (separated from the package statement). I noticed this in other classes as well

sahishnuk2 (commented):
Your coding style follows most of the coding standards. Good job 👍. Just a few areas to improve on. Keep it up!

src/main/java/Nami/Nami.java
```diff
@@ -0,0 +1,185 @@
...
+//                storage.save(tasks);
+//                continue;
+//            }
+
+//            throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+//
+//        }
+//        sc.close();
+//    }
+//}
```
mattcce: Probably should get rid of these; you can always use git to retrieve these again if you wish, instead of leaving long commented segments in the code

src/main/java/Nami/command/AddDeadlineCommand.java
mattcce: Agree with this - looks a lot cleaner with the newline

src/main/java/Nami/command/AddEventCommand.java
```diff
@@ -0,0 +1,34 @@
...
+        this.start = start;
+        this.end = end;
+    }
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        Tasks t = new Events(description, start, end);
+        tasks.add(t);
+        storage.save(tasks.asList());
+
+        System.out.println("____________________________________________________________");
```
mattcce: These lines (that decorate the output) in particular occur a lot throughout all your commands; it might be better to extract this behaviour out, so that if you ever decide to change the way your outputs are styled, you can do so easily.

src/main/java/Nami/Nami.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            String input = ui.readCommand();
+            try {
+                Command cmd = Parser.parse(input);
+                cmd.execute(tasks, ui, storage);
+                isExit = cmd.isExit();
```
mattcce: Would a `break` make more sense here, rather than using a flag?

src/main/java/Nami/task/Deadlines.java
mattcce: Agreed - it's quite jarring to see `Deadlines` or `Tasks` with only single instances.

src/main/java/Nami/task/Events.java
```diff
@@ -0,0 +1,42 @@
...
+public class Events extends Tasks{
+
+    private LocalDateTime startTime;
+    private LocalDateTime endTime;
+    public Events(String description, LocalDateTime startTime, LocalDateTime endTime) {
+        super(description);
+        this.startTime = startTime;
+        this.endTime = endTime;
+    }
+    public LocalDateTime getStartTime() {
```
mattcce, line 21: style: whitespace consistency, between members of a class (this appears to be a recurrent problem throughout the codebase)

reference: https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace

src/main/java/Nami/ui/Ui.java
```diff
@@ -0,0 +1,39 @@
+package Nami.ui;
+import java.util.Scanner;
+
+public class Ui {
+    private final Scanner scanner = new Scanner(System.in);
+
+    public void showWelcome() {
+        System.out.println("____________________________________________________________\n");
```
mattcce, line 8: `println` already inserts a newline character; no need for the `\n` (unless this was intentional; in which case please ignore)

src/test/java/Nami/parser/ParserTest.java
mattcce: Agreed except that this is also a style issue (style guide prescribes specific naming conventions for test cases), so it's less optional than suggested by previous reviewer.

mattcce (commented):
Overall, pretty good code organisation! Most issues listed are only related to style.

PR #440, opened 2025-08-28
[Glacialthorn] iP

Glacialthorn:
#JohnChatbot
> Here is a link to the Bay Harbour Butcher ([source]https://en.wikipedia.org/wiki/James_Doakes)

JohnChatbot frees your mind of having to remember things you need to do. It's,

text-based
easy to learn
~FAST~ SUPER FAST to use
All you need to do is,

download it.
double-click it.
add your tasks.
let it manage your tasks for you 😎
And it is **FREE**!

Features:

 - [X] Managing tasks
 - [ ] Reminders (coming soon)
If you are a Java programmer, you can use it to practice Java too. Here's the main method:<br/>
```
public class Main {
    }
}
```


src/main/java/JohnChatbot/JohnChatbot.java
```diff
@@ -0,0 +1,290 @@
...
+        } catch (JohnChatbotException e) {
+            Ui.printSection("Something went wrong: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Removes the chosen task from the list based on index provided.
+     * @param input The input command from the user.
+     */
+    public static void deleteTask(String input) {
```
zjaoyuki: Was this supposed to be private haha since the rest of your functions were private

src/main/java/JohnChatbot/Tasks/Event.java
```diff
@@ -0,0 +1,28 @@
...
+        super(description);
+        this.description = description;
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public String toString() {
+        if (this.isDone) {
+            String output = "[E] [X] " + this.description;
```
zjaoyuki, line 21: Just a suggestion, but perhaps can use `super`? then just add on the `[E] ` part of the toString.

src/main/java/JohnChatbot/JohnChatbot.java
```diff
@@ -0,0 +1,290 @@
...
+
+import JohnChatbot.Tasks.Deadline;
+import JohnChatbot.Tasks.Event;
+import JohnChatbot.Tasks.Task;
+import JohnChatbot.Tasks.Todo;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+public class JohnChatbot {
```
zjaoyuki, line 15: Remember to add header comment for the class as well, i.e. what this class does and a brief description of the various functions available.

src/main/java/JohnChatbot/JohnChatbot.java
```diff
@@ -0,0 +1,290 @@
...
+    private static TaskList taskList;
+    private static final String saveFileName = "JohnChatbotSave.ser";
+
+    public static void main(String[] args) {
```
zjaoyuki: I am not too sure about this but I think the main also needs a header comment (i.e. `@param args`) and what it does overall.

src/main/java/JohnChatbot/JohnChatbotException.java
```diff
@@ -0,0 +1,7 @@
+package JohnChatbot;
+
+public class JohnChatbotException extends Exception {
```
zjaoyuki, line 3: Class header comment

src/main/java/JohnChatbot/Parser.java
```diff
@@ -0,0 +1,40 @@
+package JohnChatbot;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.List;
+
+public class Parser {
+    public static final List<String> formatList = List.of("yyyy-MM-dd HHmm", "HHmm yyyy-MM-dd");
```
zjaoyuki, line 9: Have header comment for the declared variables too.

src/main/java/JohnChatbot/Parser.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final DateTimeFormatter formatter = dtFromList(formatList);
+
+    public static DateTimeFormatter dtFromList(List<String> list) {
```
zjaoyuki, line 12: Header comment for these functions as well.
Can consider standardising `dt` or `dateTime` in the function.

src/main/java/JohnChatbot/Ui.java
```diff
@@ -0,0 +1,52 @@
+package JohnChatbot;
+
+import java.util.List;
+
+public class Ui {
+    static final String lineBreak = "\n____________________________________________________________\n";
```
zjaoyuki: Since it is final, should use `LINE_BREAK` instead of `lineBreak` for constants.

src/test/java/seedu/JohnChatbot/ParserTest.java
```diff
@@ -0,0 +1,30 @@
+package seedu.JohnChatbot;
+
+import JohnChatbot.Parser;
+import org.junit.jupiter.api.Test;
```
zjaoyuki, line 4: Remember to standardise imports for every file (i.e. follow an order)
`import org.junit.jupiter.api.Test;` should either always below or above the 
`import static org.junit.jupiter.api.Assertions.assertEquals;`

zjaoyuki (commented):
Reminder to add header comments for functions, classes and declared variables >-< Overall, the code standard is there so good job but need check through all files and add header comments.

src/main/java/JohnChatbot/JohnChatbot.java
```diff
@@ -71,6 +74,28 @@ private static void run() {
...
+        }
+
+        if (matchingTasks.getTaskList().isEmpty()) {
+            Ui.printSection("No tasks match your search!");
+        } else {
+            Ui.printList(matchingTasks.getTaskList(),
+                    "Here are the tasks that match your search of " + description);
+        }
+    }
+
```
mellowfarm: The findTask method correctly filters tasks by checking if task.toString() contains the search description.

src/main/java/JohnChatbot/JohnChatbot.java
```diff
@@ -71,6 +74,28 @@ private static void run() {
...
+        String description = Parser.getFlag(input, flag);
+        if (description.isEmpty()) {
+            throw new IllegalArgumentException("Add description after 'find' to search for tasks with the word!");
+        }
+
+        for (int i = 0; i < taskList.getTaskList().size(); i++) {
+            if (taskList.getTaskList().get(i).toString().contains(description)) {
+                matchingTasks.getTaskList().add(taskList.getTaskList().get(i));
+            }
+        }
```
mellowfarm, line 118: Loops like in findTask could use enhanced for-loops for clarity

src/main/java/JohnChatbot/Tasks/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public String toString() {
+        if (this.isDone) {
+            String output = "[D] [X] " + this.description + " " + Parser.dateTimeToString(deadline);
+            return output;
+        } else {
+            String output = "[D] [ ] " + this.description + " " + Parser.dateTimeToString(deadline);
+            return output;
+        }
+    }
```
mellowfarm, line 26: The toString() method calls Parser.dateTimeToString(deadline). If the deadline is invalid or null, this could throw an exception. Currently, the exception would propagate outside this method. You might consider handling the exception here (e.g., with a try-catch) or documenting it, so that calling code doesn’t crash unexpectedly.

src/main/java/JohnChatbot/Storage.java
```diff
@@ -0,0 +1,52 @@
...
+                parentDir.mkdirs();
+            }
+
+            try (FileOutputStream fos = new FileOutputStream(file);
+                 ObjectOutputStream oos = new ObjectOutputStream(fos)) {
+                oos.writeObject(taskList);
+            }
+        } catch (IOException e) {
+            System.err.println("Error saving TaskList to file: " + filename); // The error message should print the actual filename
+        }
```
mellowfarm: Currently, exceptions like IOException or ClassNotFoundException are only printed to the console. Consider either logging them more formally or rethrowing them as a custom exception so calling code can handle errors more gracefully.

mellowfarm, line 62: In saveToFile(), the error message prints the filename but not the exception details; including e.getMessage() or e.printStackTrace() could help with debugging.

src/main/java/JohnChatbot/Parser.java
```diff
@@ -0,0 +1,33 @@
...
+        return dtFBuilder.toFormatter();
+    }
+
+    public static String dateTimeToString(LocalDateTime formatter) {
+        return formatter.format(DateTimeFormatter.ofPattern("MM dd yyyy, HHmm")) + "H";
+    }
+
+    public static LocalDateTime parseDateTime(String str) {
+        return LocalDateTime.parse(str, formatter);
+    }
```
mellowfarm, line 24: Exception may be thrown here, should have a try catch block to catch the exception and then you can throw your JohnException after with the custom error message.

mellowfarm (commented):
Overall, I think your code is pretty well written, just that you need to catch your exceptions when parsing (LocalDateTime)! Yay!

PR #480, opened 2025-08-28
[Nasif] iP

naa-siuuuu-ff:
### Ozil

>Mesut Özil made me the world's best striker - Cristiano Ronaldo

Ozil is the assist machine that assists you in remembering all your things, so that you can lock in and focus on your work.

- Super fast to use
- Interact through keyboard
-  Very gentle learning curve

And it is **FREE!** You just need a laptop!
All you need to do is.

1. Launch it
2. Give it a todo, deadline or event
3. Let it manage all the remembering for you!

Features:

- [x] Adding tasks
- [x] Adding deadlines, or timings
- [x] Finding tasks with keywoards
- [x] User interface
- [x] Arrange deadlines and events chronologically by time


src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,22 @@
+/**
+* Deadline tasks are an extension of tasks, containing a deadline
+* @param deadline Deadline of the task
+*/
+public class DeadlineTask extends Task {
+    private String deadline;
```
sj-c: Would a more descriptive name like deadlineDate or dueDate make the intent clearer (W4.6d)? Right now deadline could be mistaken for either a string description or a boolean flag.

src/main/java/Ozil.java
```diff
@@ -0,0 +1,206 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+* This is the chatbot Ozil's main class, that runs the chatbot
```
sj-c: I like the summary! Should the @param tags be moved to the constructor Javadoc instead? Class-level Javadoc usually doesn’t list parameters (W4.6c “use standard words” + correct Javadoc placement).

src/main/java/Ozil.java
```diff
@@ -0,0 +1,206 @@
...
+* @param tasks Tasks added by the user
+* @param isOpen boolean to see whether the chatbot is still running
+*/
+public class Ozil {
+    private ArrayList<Task> tasks;
+    private boolean isOpen;
```
sj-c: Name reads well for a boolean. Would isRunning better convey “main loop still active”? (W4.6d “use name to explain”.)

src/main/java/Ozil.java
```diff
@@ -0,0 +1,206 @@
...
+        this.isOpen = true;
+    }
+
+    private static void echo(String userinput) {
+        Messages.line();
+        System.out.println(userinput);
+        Messages.line();
+    }
+
+    private void addTaskToList(String userinput) {
```
sj-c: Naming: would addTodo read cleaner (W4.6e “not too long, not too short”)? The “…ToList” suffix might be redundant since we always add to the same list.

sj-c (commented):
LGTM. Some suggestions made.

src/main/java/ozil/main/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+import ozil.exception.ErrorMessages;
+import ozil.exception.OzilException;
+import ozil.task.DeadlineTask;
+import ozil.task.EventTask;
+import ozil.task.Task;
+import ozil.task.TodoTask;
+
+import java.io.*;
+
+import java.util.ArrayList;
```
lunardrak: ArrayList is imported but is not used. Maybe it could be removed

src/main/java/ozil/main/Ozil.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public void run() throws OzilException {
+        Scanner scanner = new Scanner(System.in);
+
+        Messages.intro();
+
+        Command command = new Command();
+
+        while (!command.isTerminatingCommand()) {
```
lunardrak: I like how you handled the main loop of your program!

src/main/java/ozil/main/Parser.java
```diff
@@ -0,0 +1,41 @@
...
+public class Parser {
+
+    private static boolean doesInputStartWith(String input, String command) {
+        return input.toLowerCase().startsWith(command);
+    }
+
+    public static Command handleInput(String userInput) throws OzilException {
+        String input = userInput.trim();
+
+        if (doesInputStartWith(input, "bye")) {
```
lunardrak: Perhaps this could be replaced with a switch case to reduce repeated calls of the doesInputStartWith function?

src/main/java/ozil/command/AddEventTaskCommand.java
```diff
@@ -0,0 +1,37 @@
+package ozil.command;
+
+import ozil.exception.ErrorMessages;
+import ozil.main.Messages;
```
lunardrak: Maybe the imports of ozil.main can be grouped together to make it easier to browse the list.

PR #485, opened 2025-08-28
[jsprdev] iP

jsprdev:
# Pichu — A tiny task/chat bot 🐣

> “Make it work, make it right, make it fast.” — Kent Beck ([source](https://en.wikipedia.org/wiki/Kent_Beck))

**Pichu** is a lightweight, _text-based_ assistant forked from [Duke](https://github.com/se-edu/duke) and renamed for this iP assignment. It’s **FAST**, ~~complicated~~ **simple**, and easy to extend.

### Why Pichu?
- text-based, minimal UI  
- easy to learn and hack on  
- **SUPER FAST** to use  
- friendly prompts 🤖

### Getting Started
1. Download the JAR from the **[Releases page](https://github.com/YOUR_USERNAME/ip/releases)**.  
2. Double-click it (or run via terminal with `java -jar pichu.jar`).  
3. Add your tasks.  
4. Let **Pichu** handle the rest. 😉

### Quick Demo (run locally)
```java
public class Main {
        Pichu p = new Pichu();
        p.run(); // try: list, todo Read book, deadline Submit iP /by 2025-09-30, bye
    }
}
```

### Common Commands
- `list` — show all tasks  
- `todo <desc>` — add a to-do  
- `deadline <desc> /by <YYYY-MM-DD>` — add a deadline  
- `event <desc> /from <time> /to <time>` — add an event  
- `done <index>` — mark as done  
- `bye` — exit  

You can also mark tasks as done with inline code like `done 2`.

### Features (and roadmap)
- [x] Managing tasks  
- [x] Managing deadlines  
- [ ] Reminders  
- [ ] Persistence to file  
- [ ] GUI upgrade  

> Need help? See the **[Pichu User Guide](https://github.com/YOUR_USERNAME/ip/wiki/User-Guide)**.

---

#### Notes for reviewers
- Renames the project from **Duke** to **Pichu** and completes the assignment practice tasks.  
- Includes small refactors and README updates.


src/main/java/pichu/Pichu$1.class
nty-dev, line 1: if I'm not wrong, you are not supposed to commit .class binary files. Add these to gitignore

src/main/java/pichu/Pichu.java
```diff
@@ -0,0 +1,185 @@
+package pichu;
+
+import pichu.storage.Storage;
+import pichu.core.TaskList;
+import pichu.ui.Ui;
+import pichu.parser.Parser;
+import pichu.task.Task;
+import pichu.task.ToDo;
```
nty-dev: Consider using "Todo" instead of "ToDo". This naming convention is inconsistent throughout the code. I think sticking with "Todo" makes sense.

(I'm making this comment once, but "ToDo" appears multiple more times in the code.)

src/main/java/pichu/storage/Storage.java
```diff
@@ -0,0 +1,78 @@
+package pichu.storage;
+
+import pichu.task.Task;
+
+import java.io.*;
+import java.nio.file.*;
```
nty-dev: Part of the Java Coding Standard is to not use imports with *. Import the specific module you require.

src/main/java/pichu/task/Task.java
```diff
@@ -0,0 +1,93 @@
...
+        this.name = name;
+    }
+
+    /**
+     * Sets the isCompleted field of the task.
+     *
+     * @param completed the status of completion of the task.
+     *
+     */
+    public void setCompleted(boolean completed) {
```
nty-dev, line 126: Variable name of boolean should be "isCompleted" instead of "completed" to follow the Java Coding Standard.

src/main/java/pichu/Pichu.java
```diff
@@ -0,0 +1,185 @@
...
+import pichu.task.Deadline;
+import pichu.task.Event;
+
+import java.util.List;
+import java.util.ArrayList;
+
+public class Pichu {
```
nty-dev, line 20: No JavaDoc for this public class. Can choose to add to follow the Coding Standard.

nty-dev (commented):
Nice work on this submission! The code is clean, well-structured, and follows the majority of the coding standards, which makes it easy to read and understand. Good use of consistent naming, formatting, and documentation.

One area for improvement would be to tighten consistency across the board — for example, ensuring imports, comments, and formatting strictly follow the standard, and making small adjustments to naming and encapsulation where needed. Overall, it’s a solid effort with just minor refinements to bring everything fully in line with the style guide.

src/main/java/pichu/ui/Ui.java
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Shows the welcome message to the user.
+     */
+    public void showWelcome() {
+        String name = "Pichu";
+        System.out.println("____________________________________________________________\n" +
+                " Hello! I'm " + name + "\n" +
+                " What can I do for you?\n" +
+                "____________________________________________________________\n");
+    }
```
JunhaoW3: Since horizontal line printing is used so much in your program, perhaps it is better to write a method to print horizontal lines, and simply call it at the appropriate places throughout your program.

This not only applies to this method, but to any method that requires printing of horizontal lines

src/main/java/pichu/Pichu.java
```diff
@@ -0,0 +1,185 @@
...
+            taskList.markTask(index);
+            Task task = taskList.getTask(index);
+            ui.showTaskMarked(task);
+            storage.saveAllTasks(taskList.getTasks());
+        } catch (NumberFormatException e) {
+            ui.showError("Invalid task number format.");
+        } catch (IndexOutOfBoundsException e) {
+            ui.showError("Task number is out of range.");
+        }
+    }
```
JunhaoW3: Maybe these command methods can be extracted out?

src/main/java/pichu/task/Task.java
```diff
@@ -0,0 +1,93 @@
+package pichu.task;
+
+public class Task {
+    String name;
+    boolean isCompleted = false;
```
JunhaoW3: Good variable names

JunhaoW3 (commented):
Changes can be made to extract code more and reduce repetition of rewriting code by using methods

src/main/java/pichu/Pichu$1.class
jsprdev: > if I'm not wrong, you are not supposed to commit .class binary files. Add these to gitignore

thank you for spotting that!

src/main/java/pichu/Pichu.java
jsprdev: Good suggestion, thank you!

src/main/java/pichu/ui/Ui.java
jsprdev: that is a good idea, will consider that, thank you!

PR #502, opened 2025-08-28
[TrPhuc] iP

tieutrongphuc:
### **July**
---
>“Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))
---
July not only is a chat bot, she also your best companion managing your tasks. She can: 
* Text based for ~Linux~ CLI enjoyer
* Super fast response time
>anime interface in the future ?
---
Features:
- [ ] Managing tasks todo
- [ ] Managing deadlines 
- [ ] Managing event
- [ ] Mark and unmark tasks
- [ ] Find task not done yet for a specific date
---
All you need to do is,
1. download it from [here](https://github.com/[tieutrongphuc/ip](https://github.com/tieutrongphuc/ip)).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😍
---
If you want to adjust file save path just put it in July parameter
```java
        new July("data/savefile.txt").run();
    }
```


src/main/java/July/tasks/Task.java
```diff
@@ -0,0 +1,80 @@
+package July.tasks;
+
+import July.datetime.StringTime;
+
+/**
+ * Abstract base class representing a generic task in the July task management system.
+ * This class provides common functionality for all task types including description,
+ * completion status, and basic display formatting.
+ */
+public class Task {
```
sj-c: I noticed we never actually create a plain Task — only Todo, Deadline, and Event extend it. Should this class be declared abstract instead, so that it’s clear it’s just a base type and can’t be instantiated directly?

src/main/java/July/command/keywords.java
```diff
@@ -0,0 +1,58 @@
+package July.command;
+
+/**
+ * Enumeration of valid command keywords for the July task management system.
+ * This enum defines all supported command keywords and provides functionality
+ * to parse user input and extract keywords along with their arguments.
+ */
+public enum keywords {
```
sj-c: I like that you used an enum here. Should the name be Keywords in PascalCase to follow the Java convention for enum/class names?

src/main/java/July/command/keywords.java
```diff
@@ -0,0 +1,58 @@
...
+     * This method checks if the input starts with any valid keyword and splits
+     * the input into the keyword and the remaining description/argument.
+     *
+     * @param input the user input string to parse
+     * @return a String array where index 0 contains the matched keyword (or empty string
+     *         if no match) and index 1 contains the remaining input after the keyword
+     */
+    public static String[] parse(String input) {
+        for (keywords k : values()) {
+            if (input.startsWith(k.getKeyword())) {
```
sj-c: Any reason for using startsWith here rather than splitting on the first space? Right now "listening" would also match "list". Is that intended?

src/main/java/July/command/keywords.java
```diff
@@ -0,0 +1,58 @@
...
+                String description = input.substring(k.getKeyword().length());
```
sj-c: When extracting description, would it help to also trim() it? Otherwise you might end up with a leading space if the user typed "list tasks"

src/main/java/July/tasks/Event.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Returns the save format representation of this event task.
+     * Creates a string suitable for file storage containing the task type,
+     * completion status, description, start time, and end time.
+     *
+     * @return a formatted string for saving the event task to storage
+     */
+    public String toSave() {
```
sj-c: Since this overrides a method from Task, should we add an @Override annotation here? That would help the compiler catch mismatches if the base method signature ever changes.

sj-c (commented):
LGTM. Some small fixes needed.

src/main/java/July/data/Storage.java
```diff
@@ -0,0 +1,128 @@
...
+import July.error.InvalidEventException;
+import July.error.InvalidDeadlineException;
+
+/**
+ * Write and red data to and from the file system.
+ * This class manages loading tasks from a file on startup and saving
+ * tasks to a file for persistence across application sessions.
+ */
+public class Storage {
+    private final File f;
```
nob-s: Perhaps you a more descriptive name would be better? The scope of this name is quite large

src/main/java/July/tasks/Deadline.java
```diff
@@ -0,0 +1,84 @@
...
+import July.error.InvalidDeadlineException;
+
+/**
+ * Represents a task with a specific deadline date and time.
+ * This class extends Task to include deadline functionality, allowing tasks
+ * to have a "by" date that indicates when the task should be completed.
+ */
+public class Deadline extends Task {
+
+    protected StringTime by;
```
nob-s: Perhaps a more descriptive name to show that it is of type StringTime? like byStringTime

src/main/java/July/tasks/Task.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Checks if this task is relevant for the specified date.
+     * This base implementation returns false and should be overridden
+     * by subclasses to provide appropriate date-checking logic.
+     *
+     * @param o the StringTime object representing the date to check against
+     * @return false (to be overridden by subclasses with specific logic)
+     */
+    public boolean check(StringTime o) {
```
nob-s: Perhaps could be more descriptive to what the method does?

nob-s (commented):
Overall pretty good and understandable code! Good job!

PR #528, opened 2025-08-29
[goatmilkkk] iP

goatmilkkk:
# Jay

> Your favourite task manager

Jay helps you free your mind from remembering every task. It’s,  

- **text-based**  
- *easy to learn*  
- ~~boring~~ actually **fun** to use 🎉  

All you need to do is:  

1. Download it from [this link](https://github.com/goatmilkkk/ip/releases).
2. Run it with `java -jar jay.jar`.  
3. Add your tasks.  
4. Let Jay manage them for you 😉  

## Features  

- Managing tasks  
- Managing deadlines
- Managing events

---

```java
public class Main {
        new Jay(FILE_PATH).run();
    }
}

### CODE REVIEW

File: src/main/java/jay/Storage.java [hunk 1]
```diff
@@ -71,21 +77,27 @@ public ArrayList<Task> load() throws JayException {
...
+                    if (parts.length < 4) {
+                        throw new JayException("Error, bad Deadline line.");
+                    }
                     LocalDateTime by = LocalDateTime.parse(parts[3].trim(), ISO);
                     t = new Deadline(desc, by);
                     break;
                 }
                 case 'E': {
-                    if (parts.length < 5) throw new JayException("Error, bad Event line.");
+                    if (parts.length < 5) {
```
Briescion: Perhaps you could remove the curly brace?

src/main/java/jay/Jay.java
```diff
@@ -0,0 +1,112 @@
+package jay;
+
+/**
+ * Base Chatbot application
+ */
+public class Jay {
+
+    private Storage storage;
+    private TaskList tasks;
```
Briescion: I like how the variable name is in plural to refer to a collection.

src/main/java/jay/TaskList.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Finds all tasks in the list whose description contains the given keyword.
+     * The search is case-insensitive. If the keyword is an empty string,
+     * an empty task list will be returned.
+     *
+     * @param keyword The keyword to search for in each task's description.
+     * @return A new {@code TaskList} containing the tasks that matched.
+     */
+    public TaskList findByKeyword(String keyword) {
```
Briescion: I like how you used the name to explain the purpose of the method

src/main/java/jay/Ui.java
Briescion, line 1: I like the consistent use of show in the method name as it makes it very readable

goatmilkkk: thanks for the review!

src/main/java/jay/Storage.java
```diff
@@ -0,0 +1,177 @@
...
+                    LocalDateTime from = LocalDateTime.parse(parts[3].trim(), ISO);
+                    LocalDateTime to = LocalDateTime.parse(parts[4].trim(), ISO);
+                    t = new Event(desc, from, to);
+                    break;
+                }
+
+                default: {
+                    throw new JayException("Error, unknown task type in storage: " + kind);
+                }
+                }
```
gunawanjoseph: I think the extra curly braces around each `case` block are unnecessary according to https://se-education.org/guides/conventions/java/intermediate.html

src/main/java/jay/Task.java
```diff
@@ -0,0 +1,40 @@
+package jay;
+
+/**
+ * Represents a generic task with a description and completion status.
+ */
+public class Task {
```
gunawanjoseph: I think maybe `Task` could be made `abstract` since it represents a generic concept and you never really instantiate a plain `Task` directly — only its specific types like `Todo`, `Deadline`, or `Event`.

src/main/java/jay/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+        System.out.println("\t Bye. Hope to see you again soon!");
+        System.out.println("\t____________________________________________________________\n");
+    }
+
+    /**
+     * Displays all tasks in the list.
+     *
+     * @param tasks The list of tasks.
+     */
+    public void showTasks(ArrayList<Task> tasks) {
```
gunawanjoseph: For consistency, you might want to change the parameter type from `ArrayList<Task>` to `TaskList`, since you already have a `TaskList` class.

gunawanjoseph (commented):
Great work overall! The code is clean and easy to follow with consistent naming conventions and clear Javadoc.

src/main/java/jay/Ui.java
goatmilkkk: thx for all the great comments! they were super helpful haha

PR #530, opened 2025-08-29
[Chen Chi] iP

Orange-eat97:
# [Orange_eat97] Ip
> ~Quote of the day:~

_**Orange-eat97's**_ Ip for CS2103T, AY25/26Sem1 🦔 

*this is the first line of the **Bullet List**
*and the second line

1, this is the first line of the **numbered list**
2, and here goes the second
3, maybe end with a third line?

- [x] different personality from duke?
- [ ] more personality?
- [ ] more cool features?

here we see an inline code: `public Todos()`

here we see a fenced code block:
```java
 public void run() throws EmptyTimeException, WrongHeadingException {        //gets CC bot in action
        Storage.ensureDataFileExists();
        Ui.Start();
        Ui.waitForCommand();
    }
```
canva [link](https://canvas.nus.edu.sg/) here


src/main/java/CC/CC.java
```diff
@@ -0,0 +1,23 @@
+package CC;
+
+
+public class CC {
+    private String FILE_PATH;
```
tokitobashi1: There is a minor issue here, it should not be in SCREAMING_SNAKE_CASE unless it is meant to be a constant?

src/main/java/CC/Deadlines.java
```diff
@@ -0,0 +1,20 @@
+package CC;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadlines extends Task {
+    private String Task;
```
tokitobashi1: It will be slightly confusing as you have a class Task and a private string of the same name, perhaps it should be like taskDescription instead

src/main/java/CC/Events.java
```diff
@@ -0,0 +1,24 @@
+package CC;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Events extends Deadlines {
+    private String Task;
+    private LocalDate Deadline;
+    private LocalDate Start;
```
tokitobashi1: The private fields Task, Deadline, is very similar to the name of the classes so it is slightly misleading, it is a minor issue but perhaps you should remove the capital.

src/main/java/CC/Parser.java
```diff
@@ -0,0 +1,77 @@
+package CC;
+
+import java.util.Arrays;
+
+public class Parser {
+    private static final String stop = "bye";
+    private static final String[] commands = {"todo", "event", "deadline", "mark", "unmark", "list", "delete", "find"};
+
+    public int getAction(String s) {    //controls how we parse the command
+        if (Arrays.stream(commands).noneMatch(x->s.startsWith(x))) {
```
tokitobashi1: The variable s should be slightly longer and more descriptive.

src/main/java/com/cc/Storage.java
```diff
@@ -0,0 +1,40 @@
+package CC;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class Storage {
+    private static String FILE_PATH;
```
tokitobashi1, line 13: This is a minor issue, it should not be in full caps and underscore unless it is meant to be a constant?

tokitobashi1 (commented):
Overall, the code is good, just minor cosmetic adjustments!

PR #558, opened 2025-08-29
[ZavierCSJ] iP

ZavierCSJ:
# Teemo – Swift Scout Task Manager

> “Never underestimate the power of the Scout’s code.” - Teemo

Teemo helps you keep track of tasks so you never get ganked by deadlines. It’s:

- text-based and GUI-powered  
- easy to learn  
- ~~SLOW~~ **SUPER SWIFT** to use

All you need to do:

1. [Download Teemo here](https://github.com/ZavierCSJ/ip/releases)
2. Double-click the jar or run from terminal
3. Add your tasks
4. Let Teemo ambush your to-do list for you 😎🍄

And it’s **FREE**!

Features:

- [x] Managing tasks  
- [x] Managing deadlines  
- [x] Beautiful GUI (Level-10 mushroom power!)
- [ ] Reminders (coming soon – like camouflaged Teemo)

If you’re a Java programmer, you can use Teemo to scout your Java skills. Here’s the `Launcher` method:
```
public class Launcher {
        Application.launch(Main.class, args);
    }
}
```


src/main/java/teemo/TeemoException.java
Eilyss, line 1: Would it be better to create multiple Exceptions?

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Scanner;
+import teemo.task.Task;
+import teemo.task.Todo;
+import teemo.task.Deadline;
+import teemo.task.Event;
+
+public class Storage {
+    private String filePath;
```
Eilyss: Perhaps Java Path class would provide more convenience?

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+            fileScanner.close();
+
+        } catch (IOException e) {
+            System.out.println("Error loading tasks from file: " + e.getMessage());
+        }
+        return tasks;
+    }
+
+    private Task parseTaskFromLine(String line) {
```
Eilyss: Would it be better encapsulation to have parseTaskFromLine be in the TaskList class instead?

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            String[] parts = line.split(" \\| ");
+            if (parts.length < 3) {
+                return null; // Invalid format
+            }
+
+            String type = parts[0].trim();
+            boolean isDone = parts[1].trim().equals("1");
```
Eilyss: Nice, boolean phrased like a question

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+            for (Task task : tasks) {
+                writer.write(task.toSaveFormat() + System.lineSeparator());
+            }
+            writer.close();
+            System.out.println("DEBUG: File saved successfully");
+        } catch (IOException e) {
+            System.out.println("ERROR: Failed to save file!");
+            System.out.println("ERROR: " + e.getMessage());
+            e.printStackTrace(); // This will show you exactly what went wrong
+        }
```
Eilyss: Would it be more convenient and safer to use a the try-with-resources statement instead? It eliminates the need to manually close the writer.

Details can be found here: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/teemo/TaskList.java
Eilyss, line 1: I like the naming, it is very intuitive and easy to understand.

src/main/java/teemo/Teemo.java
```diff
@@ -0,0 +1,226 @@
...
+                    throw new TeemoException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+                }
+            } catch (NumberFormatException e) {
+                ui.showError("Please enter a valid task number.");
+            } catch (TeemoException e) {
+                ui.showError(e.getMessage());
+            }
+        }
+        ui.showBye();
+        ui.close();
```
Eilyss: Would it be better to extract the case swith out from the try catch statement?

src/main/java/teemo/Ui.java
Eilyss, line 1: There seems to be a lot of repetition of show line. Perhaps an alternative way to do this would be to have a method that takes in a list of Strings, and formats it into the appropriate output? This way you only have to define one method, and instead you could call this method with multiple inputs.

src/main/java/teemo/task/Deadline.java
```diff
@@ -0,0 +1,39 @@
+package teemo.task;
+
+import java.time.DateTimeException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
+    private static final DateTimeFormatter DISPLAY_FMT = DateTimeFormatter.ofPattern("MMM dd yyyy");
+    private LocalDate by;
```
Eilyss: Perhaps it would be better to use a noun for the date? Maybe "deadline" could be a good candidate?

src/main/java/teemo/task/Event.java
```diff
@@ -0,0 +1,53 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Event extends Task {
+    private static final DateTimeFormatter INPUT_FMT =
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+    private static final DateTimeFormatter DISPLAY_FMT =
+            DateTimeFormatter.ofPattern("MMM dd yyyy h:mma");
+    private LocalDateTime from;
+    private LocalDateTime to;
```
Eilyss: Similar to my query in the Deadline class, perhaps it is better to use a noun for fields?

src/main/java/teemo/task/Task.java
```diff
@@ -0,0 +1,40 @@
...
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return isDone ? "[X]": "[ ]";
+    }
```
Eilyss: Nice! I like how you implemented this in the base Task class such that it can just be reused by its sub-classes

Eilyss (commented):
Overall, I would say that this is pretty good style wise, there are only some minor nitpicks regarding the naming of fields. In terms of structure, I do suggest maybe abstracting out the Commands from the main entry point of the project, this would help with encapsulation.

src/main/java/teemo/Parser.java
```diff
@@ -0,0 +1,116 @@
...
+     * @throws TeemoException if the deadline format is invalid or missing /by
+     */
+    public static Deadline parseDeadline(String input) throws TeemoException {
+        if (input.trim().equals("deadline")) {
+            throw new TeemoException("OOPS!!! The description of a deadline cannot be empty");
+        }
+        String taskAndDeadline = input.substring(9).trim();
+        if (!taskAndDeadline.contains("/by")) {
+            throw new TeemoException("Need a deadline! -> (deadline [description] /by [date])");
+        }
```
holeecow: I like how the parser returns specific messages for different failure modes—makes it easy to understand what went wrong. The separation between parse and model creation reads cleanly.

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+                    if (task != null) {
+                        tasks.add(task);
+                    }
+                }
+            }
+            fileScanner.close();
+
+        } catch (IOException e) {
+            System.out.println("Error loading tasks from file: " + e.getMessage());
+        }
```
holeecow: For consistency and testability, should we either rethrow as a domain exception (e.g., StorageException) or route through a logger? Printing to stdout can be noisy for consumers and hard to test.

src/main/java/teemo/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+            if (!file.exists()) {
+                file.createNewFile();
+                return tasks;
+            }
+
+            // Read existing file
+            Scanner fileScanner = new Scanner(file);
+            while (fileScanner.hasNextLine()) {
+                String line = fileScanner.nextLine().trim();
+                if (!line.isEmpty()) {
```
holeecow: isBlank() also rejects whitespace-only lines; since we already trim(), either remove trim() and use isBlank(), or keep trim() and test !line.isEmpty()—but doing both is a bit redundant.

src/main/java/teemo/TaskList.java
```diff
@@ -0,0 +1,63 @@
...
+        return tasks.get(index - 1);
+    }
+
+    public int size() {
+        return tasks.size();
+    }
+
+    public ArrayList<Task> getTasks() {
+        return tasks;
+    }
```
holeecow, line 52: Exposing the mutable internal list breaks encapsulation. Return an unmodifiable view (or a copy) and the interface type:

src/main/java/teemo/TaskList.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public boolean isValidIndex(int index) {
+        return index > 0 && index <= tasks.size();
+    }
+
+    public void markTask(int index) {
+        if (isValidIndex(index)) {
+            tasks.get(index - 1).markAsDone();
```
holeecow, line 60: The index - 1 logic is repeated in several places. Using a taskAtOneBased(...) helper may reduces duplication and centralises bounds checking.

holeecow (commented):
Thanks for the PR—functionality looks solid. My comments are coding-standards only: prefer programming to interfaces (List) and mark fields final; don’t expose mutable collections (return an unmodifiable view); centralise 1-based index checks instead of repeating index - 1; use try-with-resources with explicit UTF-8 and consider NIO (Path/Files) for I/O; avoid printing to stdout in library code (throw/log instead); remove unused imports; and add brief Javadocs (esp. to document 1-based indices). These small cleanups improve readability and robustness without changing behaviour.

PR #562, opened 2025-08-29
[Aidan Tan] iP

Chrogain:
**MelBot**
> Mel bot for your needs 😄 

Mel bot has a few features
- Text-based
- Easy to learn
- ~FAST~ _SUPER FAST_ to use

All you need to do is,

1.  download it from [here](https://github.com/nus-cs2103-AY2526S1/ip/pull/562)
2.  double-click it
3. add your tasks.
4. let it manage your tasks for you 👍 

And it is **FREE!**

Features:

- [X] Managing tasks
- [X] Managing different types of tasks
- [ ] Send reminders

## 

Here's the `main` method:

```ruby
    new Mel("data/tasks.txt").run();
}

```


src/main/java/mel/apps/Parser.java
```diff
@@ -0,0 +1,86 @@
...
+            throw new MelException.InvalidIndexException("Please input a number instead :)");
+
+        }
+
+
+        return index;
+
+    }
+
+
```
its-me-orion, line 90: I feel that the blank lines in this method can be removed, the method appear longer than it should be. This issue is also shown in other places as well from what I noticed.

src/main/java/mel/commands/DeleteCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @param tasks
+     * @param ui
+     * @param storage
+     * @throws MelException
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws MelException {
+        ui.printOut(tasks.remove(Parser.handleIndex(argument)));
+
+    }
```
its-me-orion, line 39: I realised that you have an abstract class with an abstract method "execute", is there a reason why you did not include the keyword `@Override`? Without the keyword it makes the association between this command and the abstract class Command less visible. This is also the case for other command types as well.

src/main/java/mel/commands/TaskCommand.java
```diff
@@ -0,0 +1,114 @@
...
+        } else if (task.equals("E")) {
+            String[] desc_and_time = argument.split("/from", 2);
+            if (desc_and_time.length < 2
+                    || desc_and_time[0].isEmpty()
+                    || desc_and_time[1].isEmpty()) {
+                throw new MelException.NoArgumentFoundException("event");
+
+            }
+
+            String[] fromandto = desc_and_time[1].split("/to", 2);
```
its-me-orion: This variable does not comply with the naming convention for variables. I suggest changing this to `fromAndTo` instead

src/main/java/mel/commands/TaskCommand.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param storage
+     * @throws MelException
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws MelException {
+        if (task.equals("T")) {
+            ui.printOut(tasks.add(new Todo(argument)));
+
+
+        } else if (task.equals("D")) {
+            String[] desc_and_time = argument.split("/by", 2);
```
its-me-orion: Feels like this is written by someone else? The naming convention changed snake_case which I think is not intended at your part

src/main/java/mel/exceptions/MelException.java
```diff
@@ -0,0 +1,74 @@
...
+
+            } else if (command.equals("deadline")) {
+                message = "You are missing the deadline and/or /by after \"deadline\".";
+            } else if (command.equals("event")) {
+                message = "Use the event command like this: \"(event) /from (start) /to (end)\".";
+            }
+            return message;
+
+        }
+    }
```
its-me-orion: `messager` feels like is a noun word, and thus, not appropriate for method names. I could suggest something else like `handleCommandException`

its-me-orion (commented):
Generally the structure of your code is fine and well divided. However, there is a lot of blank lines scattered throughout the code which I think can be trimmed, as well as some inconsistencies in variable naming. Other than that, the code is mostly easy to read and understand which is ideal. 👍

src/main/java/mel/apps/Ui.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * The UI class represents the interface from which the user interacts with and deals
+ * with displaying the feedback to the users.
+ */
+
+public class Ui {
+
+    private Scanner sc;
+
+    public Ui (Scanner sc) {
```
Vorld: I don't think there is an explicit rule about not having a space between the class name name and the parameters for the constructor, but perhaps it would be good practice to be consistent throughout the codebase? Elsewhere you have no space between the class name and parameters.

src/test/java/mel/commands/TaskCommandTest.java
```diff
@@ -0,0 +1,64 @@
+package mel.commands;
+
+import mel.apps.StorageStub;
+import mel.apps.UiStub;
+import mel.tasks.*;
```
Vorld, line 5: Perhaps replace the wildcard import with explicit ones?

src/main/java/mel/tasks/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public String toString() {
+        return String.format("[D]%s (by: %s)", super.toString(),
+                by.format(DateTimeFormatter.ofPattern("MMM d yyyy")));
+
+    }
+
+    @Override
+    public String toSaveString() {
```
Vorld, line 33: Good job with having names representing methods be verbs across the codebase!

src/main/java/mel/tasks/TaskList.java
```diff
@@ -0,0 +1,258 @@
+package mel.tasks;
+
+import mel.apps.Storage;
```
Vorld: Perhaps you could consider putting a space between imports from different packages to make it easier to read? Perhaps you can also review the consistency of the import ordering, sometimes mel packages are imported before java packages and sometimes its the other way around.

Vorld (commented):
Looks good! Mostly just consistency changes.

PR #579, opened 2025-08-30
[Tey Yee Siang] iP

YSTey0808:
# Amos  

> “Productivity is never an accident. It is always the result of a commitment to excellence.” – Paul J. Meyer  

**Amos** is a simple yet powerful task management application written in Java. It helps you stay on top of your todos, deadlines, and events. It is:  

- **Text-based**  
- *Easy to use*  
- ~~Confusing~~ clear and organized 😃  

All you need to do is:  

1. Download it from [here](https://github.com/your-repo-link).  
2. Run it using `java -jar Amos.jar`.  
3. Start adding your tasks.  
4. Let Amos manage your schedule for you 🚀  

### Features  
- [x] Add Todo, Deadline, and Event tasks  
- [x] Mark and Unmark tasks as done  
- [x] Delete tasks  
- [ ] Reminders (coming soon)


src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
+import java.util.List;
```
jyheng02: I think java.util.List might be unused, so it might be safe to remove

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Objects;
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Amos {
+    public static final String  LINE = "\t------------------------------------------------------------";
```
jyheng02: double space before LINE in "...String  LINE"

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+    public static Scanner scan = new Scanner(System.in);
```
jyheng02: i think scan should be set to private instead of public since only Amos should be calling scan to scan for user input, and no other class should

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+    public static ArrayList<Task> lst = new ArrayList<Task>();
```
jyheng02: same as line 8, i think it should be private for better encapsulation

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+
+    public static void echo() {
```
jyheng02: command loop uses recursion (echo() calls upon itself within the method), which might cause the call stack to grow rapidly when commands are repeated, and eventually cause stack overflow error. using a while(true) loop that breaks on "bye" might be better i think

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+        //Scanner
+        String res = scan.nextLine();
+        System.out.println(LINE);
+
+        if(Objects.equals(res, "bye")){
```
jyheng02: is there a reason you chose Objects.equals(res, "bye") instead of "bye".equals(res) for constant string comparison?

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+            //Check list
+            markAsDone(res);
+            echo();
+        } else if (res.startsWith("unmark ")){
+            //Check list
+            unmarkAsDone(res);
+            echo();
+        }  else if (res.startsWith("delete ")){
+            //Check list
+            delete(res);
```
jyheng02: can probably try to refactor the echo() call of delete to be in the loop for consistency (unmarkAsDone/markAsDone)does not call echo() from within, instead they are called from within the loop, but delete() itself calls echo() from within)

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+                int j = i+1;
+                System.out.println("\t " + j + ". " + lst.get(i));
+            }
+        }
+        System.out.println(LINE);
+        echo();
+    }
+
+    public static void markAsDone(String res) {
+        int value = Integer.parseInt(res.substring(5));
```
jyheng02: can use try/catch to throw a friendly error in case Integer.parseInt(...) throws if argument is non numeric

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+        if(value > 0 && value <= lst.size()){
+            lst.get(value - 1).markAsDone();
+        } else {
+            System.out.println("\t Cannot find such task.");
+            System.out.println(LINE);
+            return ;
```
jyheng02: extra space before semi colon, small issue 😛

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public static void unmarkAsDone(String res) {
+        int value = Integer.parseInt(res.substring(7));
+        if(value > 0 && value <= lst.size()){
+            lst.get(value - 1).unmarkAsDone();
+        } else {
+            System.out.println("\t Cannot find such task.");
+            System.out.println(LINE);
+            return ;
```
jyheng02: same as line 71

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+        System.out.println("\t OK, I've marked this task as not done yet:");
+        System.out.println("\t " + lst.get(value - 1));
+        System.out.println(LINE);
+    }
+
+    public static void add(String res) {
+        Task task;
+        if(res.startsWith("todo ")){
+            String des = res.substring(4).trim();
```
jyheng02: "todo " is of length 5, not 4 (note the spacing), so you end up relying on .trim() to remove the leading space for the description, can consider using split(" ", 2) for this which would split the user input "todo borrow book" at the first space, producing at most 2 parts so result would be ["todo", "borrow book"]. Do you think this might be better?

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+                System.out.println(LINE);
+                echo();
+                return;
+            } else {
+                task = new Todo(des);
+            }
+        } else if(res.startsWith("event ")){
+            String des = res.substring(5).trim();
+            if(des.isEmpty()){
+                System.out.println("\t OOPS!!! The description of a event cannot be empty.");
```
jyheng02: an event, small issue 😛

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+        System.out.println("\t\t" + task);
+        System.out.println("\t Now you have " + lst.size() + " tasks in the list.");
+        System.out.println(LINE);
+        echo();
+    }
+
+    public static void delete(String res) {
+        int size = lst.size();
+        String str = res.substring(6).trim();
+        int value = Integer.parseInt(str);
```
jyheng02: same parsing risk as line 65 markAsDone()

src/main/java/Amos.java
```diff
@@ -0,0 +1,167 @@
...
+            echo();
+        } else {
+            System.out.println("\t No such task to be deleted.");
+            System.out.println(LINE);
+            echo();
+        }
+
+    }
+
+    public static void main(String[] args) {
```
jyheng02: good minimal bootstrapping! 👍 👍

src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
+public class Deadline extends Task{
+    protected  final String by;
```
jyheng02: extra space between "protected" and "final", trivial

src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
...
+    public Deadline(String des){
```
jyheng02: constructor does not validate the input string, if des does not contain "/by", split("/by")[1] will throw exception.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
...
+        super(des.split("/by")[0].trim());
+        this.by = des.split("/by")[1].trim();
+    }
+
+    @Override
+    public String toString() {
```
jyheng02: looks fine, nice 👍

src/main/java/Event.java
```diff
@@ -0,0 +1,17 @@
+public class Event extends Task {
+    protected  final String from;
```
jyheng02: line 2 and 3 double spacing between "protected" and "final"

src/main/java/Event.java
```diff
@@ -0,0 +1,17 @@
...
+    protected  final String to;
+
+    public Event(String des){
```
jyheng02: constructor has no validation for "/from" and "/to", if des does not have either, exception will be thrown by split(...). consider checking presence and failing with a friendly error message 😃

src/main/java/Event.java
```diff
@@ -0,0 +1,17 @@
...
+        super(des.split("/from")[0].trim());
+        String[] parts = des.split("/from")[1].split("/to");
```
jyheng02: a bit risky to assume that des contains "/from" and "/to"

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
```
jyheng02: i think its good overall

src/main/java/Todo.java
```diff
@@ -0,0 +1,10 @@
+public class Todo extends Task{
```
jyheng02: i think this is good, except missing space between "Task" and "{"

jyheng02 (commented):
Overall I think its pretty nice work, the structure is clear and task types are consistent. To strengthen it, I’d suggest switching the recursive echo() to a while loop for safer flow, hardening parsing with split(...) and input validation to avoid crashes. Cleaning up small inconsistencies method calls will also improve readability.

src/main/java/Amos.java
sirylcereal: variable should be a noun (eg. scanner)

src/main/java/Amos.java
sirylcereal: Is "lst" considered a hard-to-pronounce variable name (based on the code quality naming standard under Week 4 topics)?

sirylcereal (commented):
Overall, naming conventions are all pretty good! 😁👍 In the entire context, the names are clear. If you want to avoid any ambiguity, you can make slight changes by making the names more full, or method names more specific. Other than those, all looks good!

PR #585, opened 2025-08-30
[jing-xiang-choong] iP

jing-xiang-choong:
# Eloise Chatbot 

**Eloise** is a lovely assistant that helps you manage your tasks. It's :

- text-based
- easy to use
- **FAST** and lightweight
- ~~boring~~ fun to interact with 

All you need to do is: 
1. Download it from [here](https://github.com/jing-xiang-choong/ip/releases).
2. Run 'java -jar Eloise.jar'.
3. Add your tasks. 
4. Let Eloise manage them for you!

---
## Features
- [X] Manage Tasks 
- [X] Set deadlines
- [X] Find by task/deadline


src/main/java/Eloise.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+
+    /**
+     * Parses command and executes it
+     * @param userInput raw command entered by users.
+     * @throws EloiseException if command is unknown or malformed
+     */
+    private static void handleInput(String userInput) throws EloiseException{
```
Nsohko: I think there should be a space before the  '{'

src/main/java/Eloise.java
```diff
@@ -0,0 +1,302 @@
...
+                            .append(items.get(i))
+                            .append(System.lineSeparator());
+                    //appends number with item then next line, using lineSeparator is better than \n
+                }
+                msgBox(list.toString().stripTrailing());
+            }
+            return;
+        }
+
+        if (lower.startsWith("mark")) {
```
Nsohko: While  not a big issue, you  may want to consider checking if it starts with "mark " (inlcuding the space). Same for the rest of the commands

src/main/java/Eloise.java
```diff
@@ -0,0 +1,302 @@
...
+
+    }
+
+
+    /**
+     * Handles command by creating a new Todo task
+     * @param userInput user inputs that begins with "todo"
+     * @throws EloiseException if description is missing
+     */
+    private static void handleToDo(String userInput) throws EloiseException{
```
Nsohko: Instead of having so many methods within this main class for each task, perhaps you could break it up into a factory method under the Task class?

src/main/java/EmptyDescriptionException.java
```diff
@@ -0,0 +1,14 @@
+public class EmptyDescriptionException extends EloiseException{
```
Nsohko: I think there could be a space before the '{' here too

src/main/java/EmptyDescriptionException.java
```diff
@@ -0,0 +1,14 @@
...
+                + "eg. " + command + " borrow book");
+    }
+}
+
+//public class EmptyDescriptionException extends EloiseException{
+//    public EmptyDescriptionException (String missing, String example) {
+//        super("oops! I can't proceed without " + missing + "\n"
+//                + "eg. " + example);
+//    }
+//}
```
Nsohko: I think this can be deleted

src/main/java/InvalidIndexException.java
```diff
@@ -0,0 +1,9 @@
+public class InvalidIndexException extends EloiseException{
```
Nsohko, line 3: Should move all related classes (e.g. exceptions) into a package

src/main/java/Eloise.java
```diff
@@ -0,0 +1,302 @@
...
+
+        while (sc.hasNextLine()) {
+            String userInput = sc.nextLine().trim();
+
+            if(userInput.isEmpty()) continue;
+
+            try {
+                handleInput(userInput);
+            } catch (EloiseException e) {
+                msgBox(e.getMessage());
```
Nsohko: I like your use of msgBox to neatly encapsulate UI functions

src/main/java/Eloise.java
```diff
@@ -0,0 +1,302 @@
...
+
+        if (lower.equals("list")) {
+            if (items.isEmpty()) {
+                msgBox("No items added yet.");
+            } else {
+                StringBuilder list = new StringBuilder();
+                for (int i = 0; i < items.size(); i++) {
+                    list.append(i + 1).append(". ")
+                            .append(items.get(i))
+                            .append(System.lineSeparator());
```
Nsohko: Good choice on using lineSeparator instead of new line

Nsohko (commented):
Overalll, very good changes. Some minor nits,

PR #594, opened 2025-08-30
[Julius Gwee Chyi Han] iP

julius-gwee:
# Bestie

> “Bye bestie~ Keep slayin and prayin!” 

This PR introduces persistent, friendly, and failure-resilient experiences for the renamed Bestie 😍  assistant while wiring the UI, storage, and parser together with graceful error handling on startup and during the command loop. Refer to the [Gradle Java plugin guide](https://docs.gradle.org/current/userguide/java_plugin.html) if you need build context while reviewing.

## Highlights

- **Durable storage** via `Storage.load` / `Storage.save` that recreates missing folders, skips corrupted records, and keeps the CLI output clean.
- **Rich temporal parsing** in `Deadline` and `Event`, powered by `DateTimeUtil`, so pretty and canonical date strings stay in sync.
- **A searchable task list**: `Parser` now routes `find` queries to `TaskList.find`, and `Ui.showFindResults` gives lightweight summaries of every match.
-
1. Type `todo <description>` to capture simple tasks instantly.
2. Use `deadline <description> /by <when>` for due items that benefit from pretty date rendering.
3. Try `event <description> /from <start> /to <end>` to track spans with dual canonical fields.

```java
// Parser snippet
case "find":
    if (parts.length < 2 || parts[1].trim().isEmpty()) {
        throw new BestieException("Please tell me what to find bestie!");
    }
    ui.showFindResults(tasks.find(parts[1].trim()));
    return false;
```

## Task list

- [ ] Persist tasks to disk through the shared `saveQuiet` helper after every mutation.
- [ ] Surface graceful loading failures with `Ui.showLoadingError` while continuing with an empty list.
- [ ] Import `java.util.List` where `TaskList.find` and `Ui.showFindResults` rely on the interface.
- [ ] Fix `DateTimeUtil.parseDate` so it returns a `LocalDate` instead of incorrectly invoking `LocalDateTime.parse`.


src/main/java/Bestie.java
```diff
@@ -0,0 +1,132 @@
...
+
+        boolean isExit = false;
+        while (!isExit) {
+            String input = sc.nextLine();
+
+            try {
+                String[] parts = input.split(" ", 2); // split input command
+                String command = parts[0];
+
+                switch (command) {
```
vanphuc1201: Should the switch and case on the same column?

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task {
+    protected String from;
+    protected String to;
```
vanphuc1201: Isn't the name is a noun like startDay and endDay better?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
+    protected String by;
```
vanphuc1201: For the field, isn't it be better for a noun? Eg: deadlineDay
I noticed the same issue in several other places too.

src/main/java/TaskType.java
```diff
@@ -0,0 +1,15 @@
+public enum TaskType {
```
vanphuc1201: I like how you use enum to deal with tasks. 💯

src/main/java/Bestie.java
```diff
@@ -0,0 +1,132 @@
...
+public class Bestie {
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>();
+
+        // Introduction
+        System.out.println("____________________________________________________________\n"
+                + " heyyy I'm Bestie\n"
+                + " whatsup?\n"
+                + "____________________________________________________________\n");
```
vanphuc1201: I like how you separate the line for readability. 👍

vanphuc1201 (commented):
LGTM. Just some nits to fix 🥇

src/main/java/Bestie.java
```diff
@@ -0,0 +1,132 @@
...
+                        + "____________________________________________________________\n");
+            } catch (Exception e) {
+                System.out.println("____________________________________________________________\n"
+                        + " Something feels off! Check your input again dear!\n"
+                        + "____________________________________________________________\n");
+            }
+        }
+    }
+
+    private static void printAdd(Task task, int count) {
```
he-yue-svg: Correct naming

he-yue-svg (commented):
I did not discover any issues with the names, good job for that!

PR #597, opened 2025-08-30
[Chang Hin Lok] iP

Hinlokk:


Puntart (commented):
Method and Class Naming looks great! :partying_face:

src/main/java/Xiaodavid/Task.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Returns the status icon of the task.
+     *
+     * @return "X" if done, " " if not done
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
+    /**
```
AppDevIn, line 41: nitpick: adding a blank line after getStatusIcon() for consistency with the spacing between other methods?

src/main/java/Xiaodavid/TaskList.java
```diff
@@ -0,0 +1,40 @@
...
+        return tasks.isEmpty();
+    }
+
+    public Task get(int index) {
+        return tasks.get(index);
+    }
+
+    public void add(Task t) {
+        tasks.add(t);
+    }
```
AppDevIn, line 56: The parameter name t works, but would task be more readable here?

src/main/java/Xiaodavid/TaskList.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class TaskList {
+    private final ArrayList<Task> tasks;
+
+    public TaskList() {
+        this.tasks = new ArrayList<>();
+    }
+
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = (tasks == null) ? new ArrayList<>() : tasks;
```
AppDevIn, line 22: I like how you handle the null check in the constructor

src/main/java/Xiaodavid/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+                throw new XiaodavidException("ehh deadline description or /by cannot be empty you goooon.");
+            }
+            return new ParsedCommand(CommandType.DEADLINE, desc, byStr);
+        }
+
+        if (cmd.startsWith("event")) {
+            if (cmd.length() <= 5 || !cmd.contains("/from") || !cmd.contains("/to")) {
+                throw new XiaodavidException("ehh an event must have description, /from and /to you goooon.");
+            }
+            String[] parts = cmd.substring(6).split("/from|/to");
```
AppDevIn, line 94: This regex split is quite clever but a comment explaining what /from|/to does would help future readers?

src/main/java/Xiaodavid/Xiaodavid.java
```diff
@@ -0,0 +1,141 @@
...
+                        var matches = tasks.findTasks(keyword);
+
+                        ui.showLine();
+                        if (matches.isEmpty()) {
+                            System.out.println("No matching tasks found.");
+                        } else {
+                            System.out.println("Here are the matching tasks in your list:");
+                            for (int i = 0; i < matches.size(); i++) {
+                                System.out.println((i + 1) + "." + matches.get(i));
+                            }
```
AppDevIn: Any particular reason for using direct System.out.println here instead of the UI class methods like the rest of the code?

src/main/java/Xiaodavid/Storage.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.*;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+
+/**
+ * Handles saving and loading of tasks from a file.
+ */
+public class Storage {
+    private String filePath;
```
AppDevIn, line 12: Would you consider making filePath final since it doesn't change after construction?

src/main/java/Xiaodavid/Ui.java
```diff
@@ -0,0 +1,83 @@
...
+            System.out.println("Here are the tasks in your list:");
+            for (int i = 0; i < tasks.size(); i++) {
+                System.out.println((i + 1) + ". " + tasks.get(i));
+            }
+        }
+        showLine();
+    }
+
+    public void showAdded(Task t, int total) {
+        System.out.println("-----------------------------------");
```
AppDevIn, line 80: Why can't you use showLine() here

PR #602, opened 2025-08-31
[balkinaveen] iP

balkinaveen:
### NovaGPT Plus

> "Unleash your full potential" - Naveen

NovaGPT frees up your mind of having to remember things you need to do. It's

- Text-based
- Easy prompts
- Fast responses

All you need to do is.
1. download it from [here](https://github.com/balkinaveen/ip/releases/download/A-Jar/NovaGPT.jar) 
2. double-click it
3. add your tasks
4. there you go! let NovaGPT manage your tasks for you

And its FREE!

- [X] Managing tasks
- [X] Find
- [X] Date support for deadline and event tasks
- [ ] Reminders (coming soon)


src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,122 @@
...
+                    "\nDo format your input: todo <task>");
+            return;
+        }
+        Task t = new Todo(text);
+        ls.add(t);
+        st.save(ls);
+        Ui.print("Got it. I've added this task:\n"
+                + t.toString() + "\nNow you have "
+                + ls.size()
+                + " tasks in the list.");
```
jvvvstin: Perhaps you may want to line break after every "\n" character instead?
```suggestion
        Ui.print("Got it. I've added this task:\n"
                + t.toString() + "\n"
                + "Now you have " + ls.size() + " tasks in the list.");
```
This may help to make your code layout similar to the expected output

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,122 @@
...
+import NovaGPT.task.Deadline;
+import NovaGPT.task.Event;
+import NovaGPT.ui.Ui;
+import NovaGPT.storage.Storage;
+import NovaGPT.exception.NovaException;
+import java.util.ArrayList;
+
+public class TaskList {
+    private static final String MARKED_MESSAGE = "Nice! I've marked this task as done:";
+    private static final String UNMARKED_MESSAGE = "OK, I've marked this task as not done yet:";
```
jvvvstin: I liked that you used variables to declare messages that you would like to print to the console. Perhaps you could consider doing this for your other commands like 'todo', 'deadline', 'event' etc. This way, their substring index can also be obtained rather than using magic literals like `input.substring(4)`
i.e.
```
private static TODO_KEYWORD = "todo";
private static TODO_SUBSTRING_INDEX = TODO_KEYWORD.length() + 1;
```
This could make it easier for future changes should certain key commands be changed.

src/main/java/NovaGPT/task/Deadline.java
```diff
@@ -0,0 +1,39 @@
...
+
+import NovaGPT.exception.NovaException;
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
+    private LocalDateTime endTimeAndDate;
+    private static final DateTimeFormatter INPUT_FORMAT = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+    private static final DateTimeFormatter OUTPUT_FORMAT = DateTimeFormatter.ofPattern("MMM dd yyyy, h:mma");
```
jvvvstin: Perhaps you could consider creating a DateTimeFormatterUtil class to handle the parsing of String datetime values and returning the DateTime in the respective patterns? This would help to reduce code duplication of the DateTime format pattern and you can just call parsing methods from this util class.

src/main/java/NovaGPT/ui/NovaGPT.java
```diff
@@ -0,0 +1,68 @@
...
+                        break;
+                    case EVENT:
+                        TaskList.handleEventTask(input, ls, st);
+                        break;
+                    case DELETE:
+                        TaskList.handleDelete(input, ls, st);
+                        break;
+                    case UNKNOWN:
+                        throw new NovaException("Hold up! I'm sorry but I don't get what that means, please try again :-(");
+                }
```
jvvvstin: Should the case blocks be indented?

jvvvstin (commented):
LGTM. I noted a few minor coding standard violations and some areas where code duplications could be prevented. 👍🏻

src/main/java/NovaGPT/task/Task.java
```diff
@@ -0,0 +1,58 @@
...
+        this.taskDescription = taskDescription;
+        isDone = false;
+    }
+
+    /**
+     * Marks a task as done
+     */
+    public void mark() {
+        isDone = true;
+    }
```
haorui2004: Just suggesting an alternative here: maybe naming it markDone/markComplete can make it easier for people to read. But the current name is also ok!

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,243 @@
...
+     * @param input The input string that the user provides
+     * @param ls ArrayList containing all saved tasks
+     * @param st Storage object handling all storage related commands
+     * @throws NovaException If the input is invalid
+     */
+    public static void handleMark(String input, ArrayList<Task> ls, Storage st) throws NovaException{
+        int listNum = Parser.parseTaskIndex(input, "mark");
+        if (listNum >= ls.size()) {
+            throw new NovaException("OOPS! Task number is out of range! Try again");
+        }
```
haorui2004: Maybe we can name listNum as index, so it looks more intuitive when we try to find a task in the taskList with a certain index

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /**
+     * Handles find Command
+     * Takes in a user input String
+     * ensures that input is valid
+     * finds all tasks containing the search String and
+     * prints a list
+     */
+    public static void handleFind(String input, ArrayList<Task> ls, Storage st) throws NovaException {
```
haorui2004: I like how most of your method names are self-explaining what they can do, that is quite effective

src/main/java/NovaGPT/storage/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+            String[] details = line.split("\\|");
+            for (int i = 0; i < details.length; i++) {
+                details[i] = details[i].trim();
+            }
+            String typeTask = details[0];
+            boolean isDone = details[1].equals("1");
+            Task t = null;
+
+            if (details[0].toUpperCase().equals("T")) {
+                t = new Todo(details[2]);
```
haorui2004: Since you have declared details[0] to be typeTask, I think you can just use it for easy reference :)

src/main/java/NovaGPT/storage/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+     * Checks if the directory and file exists
+     * creates the directory and file if required
+     * initialise an empty ArrayList
+     * loads the tasks from the specified file into the ArrayList
+     * returns the ArrayList
+     */
+    public ArrayList<Task> load(){
+        ArrayList<Task> tasks = new ArrayList<>();
+        try {
+            Path parent = filePath.getParent();
```
haorui2004: Perhaps can try name parent as directory/dir so it is more intuitive

haorui2004 (commented):
Overall I think your code quality is nice! The names for variables and methods are quite clear and descriptive. The comments are just minor suggestions. Good job!

src/main/java/NovaGPT/ui/NovaGPT.java
balkinaveen: Yes, you're right, I've fixed it. Thanks!

balkinaveen: Yes, this was a mistake. Ive fixed it. Thanks!

src/main/java/NovaGPT/storage/Storage.java
balkinaveen: yep, Ill make the changes

src/main/java/NovaGPT/command/Parser.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String KILL_SWITCH = "bye";
+
+    /**
+     * Returns the respective command based on user inputs
+     *
+     * @param input The input string that the user provides
+     * @return Command
+     */
+    public static Command parseCommandFromInput(String input) {
+        String lowerCaseInput = input.toLowerCase();
```
Huffle-Buffle: Should this handle null inputs as well? Right now it looks like it would throw a NullPointerException?

src/main/java/NovaGPT/command/Parser.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param input The input string that the user provides
+     * @param command The command string tied to the user input
+     * @return Zero-indexed integer
+     * @throws NovaException if input is invalid
+     */
+    public static int parseTaskIndex(String input, String command) throws NovaException {
+        try {
+            String numberString = input.substring(command.length()).trim();
+            int index = Integer.parseInt(numberString) - 1;
```
Huffle-Buffle: Should we check for empty numberString before parsing? Right now an empty string would cause a NumberFormatException?

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,243 @@
...
+     * Takes in a user input String, ArrayList and Storage
+     * ensures that input is valid,
+     * creates a todo task, adds it to the Arraylist and prompts ui
+     * to print the respective output message
+     *
+     * @param input The input string that the user provides
+     * @param ls ArrayList containing all saved tasks
+     * @param st Storage object handling all storage related commands
+     */
+    public static void handleTodoTask(String input, ArrayList<Task> ls, Storage st) {
```
Huffle-Buffle: I like how this handles validation and success messaging in one place. Any reason this method mixes multiple abstraction levels (parsing from raw input, domain object creation, persistence, and UI printing)? Would extracting parsing (→ Parser), construction (→ TaskFactory), and side effects (→ caller/UI) help the method follow SLAP better?

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,243 @@
...
+     * ensures that input is valid and number is within range
+     * marks the corresponding task as done and
+     * prints the respective output message
+     *
+     * @param input The input string that the user provides
+     * @param ls ArrayList containing all saved tasks
+     * @param st Storage object handling all storage related commands
+     * @throws NovaException If the input is invalid
+     */
+    public static void handleMark(String input, ArrayList<Task> ls, Storage st) throws NovaException{
```
Huffle-Buffle: I like the reuse of Parser.parseTaskIndex . Would it make sense to return a domain result here and let a higher layer handle the UI print, so the method stays at the “update domain + persist” level?

src/main/java/NovaGPT/command/TaskList.java
```diff
@@ -0,0 +1,243 @@
...
+     * ensures that input is valid,
+     * creates a deadline task, adds it to the Arraylist and prompts ui
+     * to print the respective output message
+     *
+     * @param input The input string that the user provides
+     * @param ls ArrayList containing all saved tasks
+     * @param st Storage object handling all storage related commands
+     * @throws NovaException If the input is invalid
+     */
+    public static void handleDeadlineTask(String input, ArrayList<Task> ls, Storage st) throws NovaException{
```
Huffle-Buffle: This method’s doing several things (validate, parse, construct, persist, print). Would you consider extracting parseDeadlineArgs(String) and perhaps returning a Result to shrink this method to ~5–8 lines?

Huffle-Buffle (commented):
Overall well done, just need to take note of some stuff to meet the course coding requirements! Good job !

PR #617, opened 2025-09-01
IP Progress

Butanol:
Shaniqua is a ~~_cool_~~ bot with features like:
-Todo list
-Cool name
How to bring Shaniqua to life:
1. Clone repo with `git clone`. [(innocent link)](https://www.youtube.com/watch?v=dQw4w9WgXcQ&pp=ygUXbmV2ZXIgZ29ubmEgZ2l2ZSB5b3UgdXA%3D)
2. Run her file.
Some basic commands are:
```
todo <task name>
deadline <name> /b <by when> /f <from when>
```
Todo:
- [ ] Finish tasks
- [ ] Add ascii **art** :tada:

Here's some java code:
```java
    System.out.println("Hello world");
}
```


src/main/java/TaskList.java
```diff
@@ -0,0 +1,42 @@
+public class TaskList {
+    private final Task[] tasks;
+    private int right;
+    public TaskList() {
+        this.tasks = new Task[100];
```
DonLeeHH: Perhaps using an ArrayList would be better to achieve a dynamic list size of tasks

src/main/java/TaskList.java
```diff
@@ -0,0 +1,42 @@
...
+        tasks[right] = x;
+        right++;
+    }
+    public void list() {
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < right; i++) {
+            System.out.printf("%d. %s", i + 1, tasks[i]);
+        }
+    }
+    public boolean mark(int idx) {
```
DonLeeHH: Following the coding standard, "Boolean variables/methods should be named to sound like booleans", I noticed this issue in other areas as well. 

I saw that your main logic uses this to check if the task was successfully marked. 
Perhaps using some exception throwing/handling?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,42 @@
+public class TaskList {
+    private final Task[] tasks;
```
DonLeeHH: I like how you used plural form to represent a collection of tasks

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
+public class Task {
+    private boolean done;
+    private final String name;
+    public Task(String name) {
+        this.name = name;
+        this.done = false;
+    }
+    public String toString() {
```
DonLeeHH: Perhaps include the @Override annotation here. Noticed this in Deadline and Event class as well

src/main/java/Shaniqua.java
```diff
@@ -0,0 +1,75 @@
...
+                String[] params = Shaniqua.handleDeadline(commands[1]);
+                tasks.addTask(new Deadline(params[0], params[1]));
+            } else if (commands[0].equals("event")){
+                String[] params = Shaniqua.handleEvent(commands[1]);
+                tasks.addTask(new Event(params[0], params[1], params[2]));
+            }
+        }
+    }
+
+    /**
```
DonLeeHH: I like the use of JavaDocs

src/main/java/Deadline.java
```diff
@@ -0,0 +1,10 @@
+public class Deadline extends Task{
+    private String by;
```
ruissouw: perhaps a better variable name would be dueBy?

src/main/java/Event.java
```diff
@@ -0,0 +1,12 @@
+public class Event extends Task{
+    private String from;
+    private String to;
```
ruissouw: likewise, perhaps something like startDate and endDate?

src/main/java/Shaniqua.java
```diff
@@ -0,0 +1,75 @@
+import java.util.Scanner;
+
+public class Shaniqua {
+    public static void main(String[] args) {
+        TaskList tasks = new TaskList();
+        Scanner input = new Scanner(System.in);
```
ruissouw: perhaps scanner or sc would be a better name? reason being the scanner itself is not the input but rather an object that renders the next input

src/main/java/Shaniqua.java
```diff
@@ -0,0 +1,75 @@
...
+        System.out.print("Kia Ora! I'm Shaniqua!\nWhat can I do for you?\n");
+        while (true) {
+            String in = input.nextLine();
```
ruissouw: is in a shortform for input? if the scanner object is renamed to scanner, then that frees up the input name for this

src/main/java/Shaniqua.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     *
+     * @param param an array of strings
+     * @return String without the first argument
+     */
+    private static String[] handleInput(String param) {
+        String[] temp = param.split(" ");
+        StringBuilder x = new StringBuilder();
```
ruissouw: perhaps sb would be a more intuitive name here

src/main/java/Shaniqua.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        return new String[]{temp[0], x.toString(), };
+    }
+    private static String[] handleDeadline(String param) {
+        return param.split("/by");
+    }
+    private static String[] handleEvent(String param) {
+        return param.split("/to|/from");
+    }
+    private static boolean integerCheck(String param) {
```
ruissouw: maybe checkInteger is a better name since it is a verb?

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
+public class Task {
+    private boolean done;
```
ruissouw: isDone would be a better name here for a boolean variable

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+    public String toString() {
+        return String.format("[%s] ", done ? "X" : " ") + name;
+    }
+    public void mark() {
+        this.done = true;
+    }
+    public void unmark() {
+        this.done = false;
+    }
```
ruissouw: instead of plainly saying mark and unmark, might it be more informative to say markAsDone and unmarkAsDone?

ruissouw (commented):
Namings are generally ok, but need to be careful about boolean variable names and some method names. Great effort though!

PR #623, opened 2025-09-01
[Ian Chin] iP

ianchinn:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,2 @@
+package PACKAGE_NAME;public class Deadline {
```
noahseethorcodes: According to the style guide, should
``` java
public class Deadline {
```

be on a new line?

src/main/java/Event.java
```diff
@@ -0,0 +1,2 @@
+package PACKAGE_NAME;public class Event {
```
noahseethorcodes: According to the style guide, should
``` java
public class Event {
```

be on a new line?

src/main/java/Todo.java
```diff
@@ -0,0 +1,2 @@
+package PACKAGE_NAME;public class Todo {
```
noahseethorcodes: According to the style guide, should
``` java
public class Todo {
```

be on a new line?

src/main/java/Yapper.java
```diff
@@ -0,0 +1,6 @@
+public class Duke {
```
noahseethorcodes: Do you intend
``` java
Duke
```

to be named

``` java
Yapper
```

instead, as indicated by the file name?

noahseethorcodes (commented):
Looks like a good start! Just bear in mind the style guide so you don't need to reformat too much down the line.

src/main/java/Deadline.java
darrensimmx: Perhaps public class Deadline should be in a new line?

src/main/java/Event.java
darrensimmx: Perhaps public class Event should be on a new line?

src/main/java/Todo.java
darrensimmx: Perhaps public class Todo should be on a new line?

src/main/java/Yapper.java
darrensimmx: Perhaps you mean to name it Yapper instead of Duke, as shown in your file name?

darrensimmx (commented):
Clean Code, however there is room for improvement

PR #627, opened 2025-09-01
[Retribob] iP

Retribob:
# **Bobby**
Get a neat little task manager in the form of *Bobby*.

Bobby is:
- Fast
- Saves your tasks for easy access
- Can manage deadlines!

To start using Bobby:
1. Download it from [here](https://github.com/Retribob/ip/releases/tag/A-Jar)
2. Open the jar file
3. Add tasks of different types using todo/deadline/event
4. End Bobby using the command bye
5. Tasks are saved on exit! 😄 

> God gave you hands to use Bobby - Anonymous

Currently implemented:
- [x] Adding and displaying tasks
- [x] Adding different types of tasks
- [ ] Supporting different time formats 

Feel free to look through the code and make changes if needed. Here the `Bobby` class:

```
public class Bobby {
    private static boolean isRunning;
    private Scanner scanner;
    private ListManager listManager;
    private Parser parser;
    private UI ui;

    /**
     * Sets isRunning to true.
     * Initializes ListManager, parser and ui instances.
     */
    public Bobby() {
        isRunning = true;
        parser = new Parser();
        listManager = new ListManager();
        ui = new UI();
    }

    /**
     * Sets boolean isRunning to false.
     * Closes UI and ListManager instances.
     */
    public void endChat() {
        listManager.closeList();
        ui.onEnd();
        isRunning = false;
    }

    private void run() {
        while (isRunning) {
            try {
                boolean isAction = parser.parseInput(listManager);
                ui.printLine();

                if (!isAction) {
                    endChat();
                }
            } catch (Exception e) {
                System.out.println(e.getMessage());
            }
        }
    }

        Bobby chatBot = new Bobby();
        chatBot.run();
    }
}
```


0rbita1 (commented):
LGTM. 

All public methods and classes have appropriate JavaDoc comments. Very good separation of concerns with the use of classes and packages. OOP principles are abided to strongly. Could not find any coding style violations.

src/main/java/uiManager/UI.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    private void onStart() {
+        System.out.println(new StartText().getText());
+        printLine();
+    }
+
+    public void onEnd() {
+        System.out.println(new EndText().getText());
+    }
```
JG-Kiran: Since they are still actions, perhaps more verb-like name, such as startChat, endChat?

src/main/java/listManager/Event.java
```diff
@@ -0,0 +1,80 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Subtype of <code>Task</code>, aside from a task name
+ * it also has a start and end date
+ */
+public class Event extends Task{
+    private String start;
+    private String end;
```
JG-Kiran: Might be confusing as nouns or verbs. How about startasString, or perhaps just startTime to distinguish itself from startDate? Same goes for end.

src/main/java/listManager/ListManager.java
```diff
@@ -0,0 +1,156 @@
...
+                            + task.getTaskWithStatus());
+    }
+
+    /**
+     * Deletes the task at the specified index.
+     *
+     * @param index The task number in the list.
+     * @throws NoSuchTaskException If index > taskList.size().
+     */
+    public void deleteTasks(int index) throws NoSuchTaskException{
```
JG-Kiran: Since you're deleting one task at a time, it should be named as 'deleteTask'?

JG-Kiran (commented):
Overall I think the naming is very clear, with little to no room for confusion. There are some instances where the distinction between fields or that between nouns and verbs could have been made clearer. Besides, the naming conventions, the spacing can be standardised. Particularly the spaces before the { at the start of class or method codes, after the signatures.

src/main/java/listManager/ListManager.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param task task description in string format.
+     * @throws NoSuchTaskException If task description doesn't match any known format.
+     * @throws IncompleteTaskException If task description matches known format but is incomplete.
+     */
+    public void add(String task) throws NoSuchTaskException, IncompleteTaskException{
+        Task taskType = taskClassifier(task);
+        taskList.add(taskType);
+        taskCount += 1;
+        uncompletedTasks++;
+        System.out.println("Task Added: " + taskType.getName()
```
isaactoh: Would it be better to move printing to the terminal to the UI class?

src/main/java/listManager/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    protected String taskDescriptor;
+    protected boolean isComplete;
+    protected String taskName;
+
+    public Task(String taskDescriptor) {
+        this.taskDescriptor = taskDescriptor;
+    }
+
+
+    public String toStringFormat() {
```
isaactoh: Why not override the toString method of Object as it is more widely understood?

src/main/java/listManager/Todo.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+    /**
+     * Processes taskDescriptor and splits it into task name.
+     *
+     * @param taskDescriptor String of user input passed into constructor.
+     * @return Task name in string format.
+     * @throws IncompleteTaskException If taskDescriptor is in known format but incomplete.
+     */
+    public String descriptorProcessor(String taskDescriptor) throws IncompleteTaskException {
```
isaactoh: As this is a method, would a processDescriptor (verb) be better than a noun?

PR #632, opened 2025-09-01
[Angmar2722] iP

Angmar2722:
# Focus

> “Focus is not about saying yes to everything. It’s about saying no to most things.” – Steve Jobs

Focus is a **no-nonsense task manager** built to help you cut through the clutter and keep track of what matters.  
Unlike bulky apps, Focus stays out of your way and gets the job done.

---

## Why Focus?
- Minimalist by design  
- *Reliable* even for long task lists  
- ~~Overwhelming features~~ → **Straightforward essentials**  
- Keyboard-first 🚀

---

## Quick Start
1. [Grab the release here](https://github.com/Angmar2722/ip/releases)  
2. Run it: `java -jar focus.jar`  
3. Type your first command (e.g., `todo read CS2103T notes`)  
4. Watch your productivity grow 📈

---

## Core Features
- Add **todos**, **deadlines**, and **events**  
- Mark ✅ and unmark ❌ tasks  
- JavaFX interface 

---

## Example
Here’s a peek at the entry point:

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Event.java [hunk 1]
```diff
@@ -0,0 +1,17 @@
+public class Event extends Task{
```
ongyixin: Should there be a space before the constructor parenthesis?

src/main/java/Focus.java
```diff
@@ -0,0 +1,128 @@
...
+
+        while (true) {
+
+            String input = scanner.nextLine();
+
+            // Split into command + rest. Keep spaces in the rest.
+            String[] parts = input.split(" ", 2);
+            String cmdString = parts[0];
+            Command cmd = Command.fromString(cmdString);
+            String taskParams = (parts.length > 1) ? parts[1] : ""; //For user input with 2 or more words
```
ongyixin: Good descriptive name `taskParams`! Would you consider extracting command handlers such as `handleToDo(String)` so that the `switch` becomes more readable?

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
...
+ * Represents a task
+ */
+
+public class Task {
+
+    protected String description;
+    protected boolean isDone;
+
+    /**
+     * Initializes task with task description and whether the task is done
```
ongyixin, line 16: Should this be rephrased to "initializes... tasks not done yet" since it sets `isDone = false`?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,55 @@
...
+    public void deleteTask(int i) {
+        String deletedTaskString = this.tasks.get(i).toString();
+        this.tasks.remove(i);
+        System.out.println("     Noted. I've removed this task:");
+        System.out.printf("       %s\n", deletedTaskString);
+        System.out.printf("     Now you have %d tasks in the list.\n", this.tasks.size());
+        printLine();
+    }
+
+    public void addTask(Task task) {
```
ongyixin: Methods like `printTaskList()`, `markTaskAsDone()`, `addTask()` both mutate and print. Consider only mutating using methods like `addTask(Task task)` for clarity and specificity in behaviour? Would it be good if you add a separate UI class to deal with user interactions?

src/main/java/FocusException.java
```diff
@@ -0,0 +1,10 @@
+/**
+ * Unchecked error for Focus command handling.
+ */
+public class FocusException extends RuntimeException {
```
ongyixin: Why did you choose to extend `RuntimeException` instead of `Exception`?

ongyixin (commented):
Hi! I'm your assigned PR reviewer. Your code is readable and logical, with consistent and intentional naming conventions. My most significant comment is to separate printing from logic, for greater clarity in method and class names. Good job!

src/main/java/FocusException.java
Angmar2722: Thank you for pointing this out. Resolved!

src/main/java/TaskList.java
Angmar2722: Added a separate UI class. Thank you!

PR #641, opened 2025-09-02
[WeeMee03] iP

WeeMee03:
**ELENA**

Elena Bot helps you manage tasks so that you don’t have to keep them all in your head. It’s,
- text-based
- easy to learn
- ~~tedious~~ SUPER FAST to use 🚀

All you need to do is:
1. Download it from [here](https://github.com/WeeMee03/ip/releases)
2. Double-click it.
3. Type in your tasks.

Let Elena handle the rest 😉

Features

- [x] Manage to-do tasks
- [x] Manage deadlines
- [x] Manage recurring tasks

```
public class Main {
    }
}
```

Elena Bot is free and fun to use ✨


dingf3ng (commented):
Looks good to me. Good OOP. Maybe abstracts IO can result in better code quality :)

williamoose (commented):
Great job on complying with code naming guidelines! Overall I couldn't find any improvements for code naming.

src/main/java/elena/AddDeadlineCommand.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param tasks   the task list to update
+     * @param ui      the user interface for displaying messages
+     * @param storage the storage for saving tasks
+     * @throws ElenaException if input format is invalid
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws ElenaException {
+        String content = input.substring(9).trim();
+        String[] parts = content.split("/by", 2);
```
DJsudartha: parts is unclear. Maybe consider changing it into deadlineData

src/main/java/elena/Elena.java
```diff
@@ -0,0 +1,240 @@
...
+                printError(e.getMessage());
+            } catch (Exception e) {
+                printError("Something went wrong: " + e.getMessage());
+            }
+        }
+
+        scanner.close();
+    }
+
+    private static void printWelcomeMessage() {
```
DJsudartha: I think all of this can be seperated to a seperate class called UI

src/main/java/elena/ElenaException.java
```diff
@@ -0,0 +1,32 @@
+package elena;
```
DJsudartha: Great exception class, very clear!

PR #648, opened 2025-09-03
[learningcoding-pixel] iP

learningcoding-pixel:


src/main/java/cat/Ui.java
```diff
@@ -0,0 +1,18 @@
+package cat;
+
```
DavidShenJingtao, line 2: How about adding javaDoc to your code?

subhitsha-sp (commented):
LGTM! The naming follows the standard conventions.

text-ui-test/EXPECTED.txt
aliensarefake, line 1: LGTM – perhaps you can add this file to your .gitignore

src/main/java/cat/Exception.java
```diff
@@ -0,0 +1,19 @@
+package cat;
+
+public class Exception extends java.lang.Exception {
+
+    // Constructor for the exception
+    public Exception(String message) {
+        super(message);
+    }
```
DavidShenJingtao, line 8: What about implementing more types of exceptions to handle different errors?

src/main/java/cat/Meow.java
```diff
@@ -0,0 +1,114 @@
...
+                    tasks.add(todo);
+                    printAll(tasks);
+                }else if(userInput.contains("deadline")) {
+                    int index = userInput.indexOf('/');
+                    name = userInput.substring(9, index);
+                    String endDate = userInput.substring(index + 1);
+                    Deadline deadline = new Deadline(name, endDate);
+                    tasks.add(deadline);
+                    printAll(tasks);
+                } else if(userInput.contains("event")) {
```
uylulu: Should this be better formatted?

src/main/java/cat/Event.java
```diff
@@ -0,0 +1,51 @@
...
+     * Example: [E][X] Meeting (Start: 01/09/2025 1400) (End: 01/09/2025 1600)
+     *
+     * @return A string representing the task's completion status along with start and end dates.
+     */
+    @Override
+    public String getStatus(){
+        String stat=  isDone ? "[X]" : "[ ]";
+        return "[E]" + stat + " " + name + " " + "(Start: " + startDate + ") " + "(End:" + endDate + ")" ;
+    }
+
```
uylulu, line 48: I like how easy it is to understand your java doc

src/main/java/cat/Dataloader.java
```diff
@@ -0,0 +1,69 @@
+package cat;
+
+import java.io.*;
+import java.util.*;
```
uylulu, line 4: Maybe it is better to import only what you need.

PR #668, opened 2025-09-04
Loh Zhi Qin iP

lohzq88:
Skye
> The bot for your task tracking

A new instance of `Skye` can be generated by
```
Skye inst = new Skye();
```
You may _download_ the source code from [here](https://github.com/lohzq88/ip/releases). And it is **FREE**.

Some features we have developed
- [X] Set events
- [X] Search list of events
- [X] Delete events


src/main/java/Skye.java
```diff
@@ -0,0 +1,111 @@
...
+        String logo = " ____  _               \n"
+                + "/ ___|| | ___   _  ___ \n"
+                + "\\___ \\| |/ / | | |/ _ \\\n"
+                + " ___) |   <| |_| |  __/\n"
+                + "|____/|_|\\_\\\\__, |\\___|\n"
+                + "            |___/      \n";
+        System.out.println("Hello from\n" + logo);
+
+        while (scanner.hasNextLine()) {
+            String line = scanner.nextLine();
```
pekyiheng: You could rename this to userInput instead to make it more clear

src/main/java/Skye.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+            } else if (line.startsWith("todo")) {
+                String description = line.split(" ", 2)[1];
+                inst.addTask(new ToDo(description));
+            } else if (line.startsWith("deadline")) {
+                String details = line.split(" ", 2)[1];
+                String[] content = details.split(" /by ");
+                inst.addTask(new Deadline(content[0], content[1]));
+            }
+            else if (line.startsWith("event")) {
```
pekyiheng: else if statement should be one line up to follow style of previous else if statements

src/main/java/Skye.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Exits the program
+     */
+    public static void exitProgram() {
+        System.out.println("Goodbye. Hope to see you again soon!");
+        System.exit(0);
+    }
+
+    public static void main(String[] args) {
+        Skye inst = new Skye();
```
pekyiheng: You could rename this to instanceSkye

src/main/java/Skye.java
```diff
@@ -0,0 +1,111 @@
...
+                String[] content = details.split(" /by ");
+                inst.addTask(new Deadline(content[0], content[1]));
+            }
+            else if (line.startsWith("event")) {
+                String details = line.split(" ", 2)[1];
+                String description = details.split(" /from ")[0];
+                String[] time = details.split(" /from ")[1].split(" /to ");
+                inst.addTask(new Event(description, time[0], time[1]));
+            }
+        }
```
pekyiheng: What if the user enters a command that is not recognised? I think there should be an else statement to handle that case

src/main/java/Skye.java
```diff
@@ -45,7 +47,7 @@ public void markTask(int number) {
      */
     public void unmarkTask(int number) {
         printString("Ok, I have marked the task below as incomplete:");
```
TeeJinSeng: For the sake of consistency with `markTask`, can move this line of printString 2 lines below, also printString should be done after the process is finished.

src/main/java/Skye.java
```diff
@@ -90,8 +92,19 @@ else if (line.startsWith("mark")) {
                 } catch (IndexOutOfBoundsException | NumberFormatException e) {
                     System.out.println("Invalid task number!");
                 }
-            } else {
-                inst.addTask(line);
+            } else if (line.startsWith("todo")) {
```
TeeJinSeng: Can I have `todosdescription` here? Does your code treat it as invalid or as a valid command to create ToDo?

src/main/java/ToDo.java
```diff
@@ -0,0 +1,11 @@
+public class ToDo extends Task{
```
TeeJinSeng: need a space between "... extends Task" and "{"

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
```
TeeJinSeng: need a space between "... extends Task" and "{"

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task{
```
TeeJinSeng: need a space between "... extends Task" and "{"

src/main/java/Task.java
```diff
@@ -3,24 +3,24 @@ public class Task {
     protected boolean isComplete;
 
     public Task(String name) {
-        taskName = name;
-        isComplete = false;
+        this.taskName = name;
+        this.isComplete = false;
```
TeeJinSeng: Actually in this course, it is recommended to avoid unnecessary `this` keyword, you can refer to this: 
https://se-education.org/guides/conventions/java/index.html#variables

TeeJinSeng (commented):
Overall is still quite good, just that there are some minor mistakes and might need to handle more possible user input error that need to be fixed.

PR #674, opened 2025-09-05
[syemfai] iP

syemfai:


src/main/java/Garfield.java
```diff
@@ -1,14 +1,31 @@
...
+            if (input.equals("bye")) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
+                break;
+            }
+
+            System.out.println("____________________________________________________________");
+            System.out.println(" " + input);
+            System.out.println("____________________________________________________________");
```
TzHsien: To save time and effort, maybe you could create a Box function that automatically wraps your string output/input in these horizontal lines?

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task {
+    protected String from;
+    protected String to;
```
TzHsien: Interesting use of protected visibility, but I believe a better practice would be to make it private, since we do not want to accidentally change the attributes of objects (other than through getters and setters)?

src/main/java/Garfield.java
```diff
@@ -60,12 +60,49 @@ public void run() {
...
+                System.out.println(" Now you have " + items.size() + " tasks in the list.");
                 System.out.println("____________________________________________________________");
+            } else if (input.startsWith("deadline ")) {
+                try {
+                    String[] parts = input.substring(9).split(" /by ");
+                    String desc = parts[0].trim();
+                    String by = parts[1].trim();
+                    Task task = new Deadline(desc, by);
+                    items.add(task);
+                    System.out.println("____________________________________________________________");
```
TzHsien: Instead of using println, maybe you consider using your boxing function that you created?

TzHsien (commented):
You are on the right track!! I see you have yet to commit the branches from week 3's iP, don't be disheartened you got this!

src/main/java/Event.java
dingf3ng: Seems private is better. As per the requirements of java style guide.

src/main/java/Garfield.java
```diff
@@ -0,0 +1,163 @@
...
+            System.out.println(" Noted. I've removed this task:");
+            System.out.println("   " + task);
+            System.out.println(" Now you have " + items.size() + " tasks in the list.");
+            System.out.println("____________________________________________________________");
+        } catch (Exception e) {
+            throw new GarfieldException("OOPS!!! Invalid task number for delete.");
+        }
+    }
+
+    private void handleUnmark(String input) throws GarfieldException {
```
dingf3ng: Seems more abstractions on UI is needed. But you are on the right track, just few more works to shrink the main class.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
```
dingf3ng: Again, use private seems to be more appropriate.

PR #706, opened 2025-09-08
refactor: Add Assertions

grenn24:
This PR introduces Java assert statements in the core classes to enforce internal invariants and catch potential programming errors during development.

Notes:
- Assertions are disabled by default and should not replace normal exception handling for user errors.
- Existing exception handling for invalid user input remains unchanged.


grenn24:
45y

